PR #1, opened 2019-10-11
Simpler and faster checking for embed indices

caryan:
Some of the commits are gratuitous style choices but it is much faster

Master
```julia
julia> indices = [3, [4,5]]
2-element Array{Any,1}:
 3      
  [4, 5]

julia> @btime QuantumOpticsBase.check_embed_indices($indices)
  1.401 μs (26 allocations: 1.95 KiB)
```

This PR
```julia
julia> @btime QuantumOpticsBase.check_embed_indices($indices);
  505.733 ns (15 allocations: 848 bytes)
```


david-pl:
Cool, thanks!

PR #15, opened 2021-03-29
Phase space functions declaration

myrddin89:
It would be useful for other packages to be able to extend functions of QuantumOptics without depending on QuantumOptics.jl itself. I added to QuantumOpticsBase.jl declarations of phase space functions: `qfunc`, `wigner`, `coherentspinstate`, `qfuncsu2`, `wignersu2`.
I especially needed `wigner` for a package I am developing, but also the other functions may be useful to others. I made changes also to QuantumOptics.jl to accommodate this. 
I hope this can be useful.


myrddin89:
Corresponding pull request in qojulia/QuantumOptics.jl#297.

david-pl:
> It would be useful for other packages to be able to extend functions of QuantumOptics without depending on QuantumOptics.jl itself.

I agree in principle, but it kind of brings up the question where we stop. For example, it won't make a whole lot of sense to define functions for `timeevolution` in QOBase. I suppose we can just move things around when someone needs it (like you did here). Thanks btw.

PR #18, opened 2021-04-11
Relaxed type constraint in DenseOperator constructor

### LINKED ISSUE #17: Method dense() doesn't preserve element type
Is there a specific reason that calling dense(x::AbstractOperator) forces conversion to ComplexF64 data element type?

Maybe this is just left over from when all abstract operators were implemented as ComplexF64, or maybe I'm missing something, but to me it would make more sense to implement dense as

`dense(x::AbstractOperator) = Operator( x.basis_l, x.basis_r, Array(x.data) )`

since this would preserve the type of the data field. This would also be more consistent with the sparse method which does preserve data type.

sd109:
Fixes #17 and a similar bug in `exp(::DenseOperator)` method which converted real operators to complex with zero imaginary part.


sd109:
This now breaks the `normalize!(::Operator)` method for integer data types due to the use of `rmul!(op.data, 1.0/tr(op))` in the normalize definition. Can we just change this method to `normalize!(op::Operator) = op / tr(op)`? Is there a particular reason that this is implemented using `rmul!`?

sd109:
This now fails `normalize!` test because `rmul!(op.data, 1.0/tr(op)` fails when op has integer data type. I don't know enough about the inplace linear algebra functions to know what the best way to fix this is.

sd109:
It turns out that `float()` handles all conversions sensibly (including things like converting `Complex{BigInt}` to `Complex{BigFloat}`) so that solves the integer problem by converting to the most sensible corresponding float type. Now there just seems to be some printing error that needs fixed.

david-pl:
Thanks for this!

Why didn't you stick with the first version though? The error with `Int`s is just how Julia handles this, e.g. `normalize!([1,1])` will throw the same error, so it should be fine. The test will pass if you explicitly change the datatype of the operator there to floats.

Actually, I also saw that while `sparse` conserves the datatype, the `SparseOperator` method doesn't. Would you mind doing that one as well? It's here https://github.com/qojulia/QuantumOpticsBase.jl/blob/201b6305440b456b4b1936356e82735cfe6b2f90/src/operators_sparse.jl#L17

sd109:
You're right that it makes sense for `normalize!` to fail for integers if this is the default Julia behaviour. Thanks for spotting the `SparseOperator` issue, I've corrected that now too.

These fixes cause a couple of small issues in the `QuantumOptics.timeevolution` module when trying to use a real Hamiltonian. I will open a PR over on QuantumOptics.jl later to fix those issues, but in the mean time it's probably better not to merge this just yet.

david-pl:
> These fixes cause a couple of small issues in the QuantumOptics.timeevolution module when trying to use a real Hamiltonian.

The test for QuantumOptics with this change pass for me locally, at least. So I hope there are no serious issues. It might well be that some type-promotion gets the wrong output type when calling something with real Hamiltonians since everything used to be hard-coded on `ComplexF64`, so no guarantees there.

> I will open a PR over on QuantumOptics.jl later to fix those issues, but in the mean time it's probably better not to merge this just yet.

Perfect, thanks a lot :+1:

PR #19, opened 2021-04-16
Fix documentation of `negativity`

KKleinbeck:
The documentation doesn't fit the code; which implements: https://en.wikipedia.org/wiki/Negativity_(quantum_mechanics)


david-pl:
Yup, you're right, thanks!

Issue #24, opened 2021-05-12
Typed methods to construct states


myrddin89:
I came across this issue while trying to make QuantumOptics interoperate with different data types. I’m not sure what should be the best approach at the moment. I see that `basisstate` accepts also a keyword `dType` to specify the data type, but this is not propagated in other places. 

I think that the Julian way is to define methods accepting also a type, usually as the first argument, with default methods when the type is not given. I mean something like:

```julia
fockstate(::Type{T}, b::FockBasis, n::Integer) where {T <: Number} = …
fockstate(b::FockBasis, n::Integer) = fockstate(ComplexF64, b, n)
```

If you think this can be useful, we can discuss the best approach to implement this, as I do not know which impact this can have and how much effort is needed to implement it.

david-pl:
The `dType` kwarg was a half-hearted approach to this. I like the signature you suggest as it does seem a lot more like idiomatic Julia.

Doing this just for states, I think we'd need to do implement methods for:
- `fockstate`
- `spinup`, `spindown`
- `nlevelstate`
- `randstate`
- `basisstate`
- `Ket` and `Bra` constructors?

Did I miss any? Also, we should probably deprecate the `dType` kwarg.

The other question is that, if we're doing this for states, do we want such methods for operators as well? That might be a bit more effort.

myrddin89:
Wow you already did the changes, thanks! I looked into the PR and seems fine. Only one thing, why did you kept the type generic: I mean, shouldn’t the type be a subtype of Number? Are there any reasons to keep it more general?

myrddin89:
Regarding your last point, I think that eventually this should be done also for operators, depending on how they are created (I admit I used little of all the QuantumOptics.jl ecosystem). If you can point me in the right direction, I can do the changes myself.

david-pl:
> Only one thing, why did you kept the type generic?

My take on this is that types should only be put there if really needed to keep things as generic as possible. If the provided type is `<:Number` then this doesn't matter at all so why not leave it out. As a side note: I got is wrong in a *lot* of places in QuantumOptics.jl. Changing type-signatures that are too restrictive will be the subject of another (rather tedious) PR.

> If you can point me in the right direction, I can do the changes myself.

[...]

PR #26, opened 2021-05-14
Added typed construtors for operators FockBasis

myrddin89:
I initiated to apply the changes as discussed in #24. Before completing I think we should review the changes I made so far. I think there are some options also for the docs. The changes I made in `fock.jl` for the operators are what I am used to. I can modify, if you consider to be better another way.

### DIFF
--- a/src/fock.jl
+++ b/src/fock.jl
@@ -21,44 +21,63 @@ end
 ==(b1::FockBasis, b2::FockBasis) = (b1.N==b2.N && b1.offset==b2.offset)
 
 """
-    number(b::FockBasis)
+    number([T=ComplexF64,] b::FockBasis)
 
-Number operator for the specified Fock space.
+Number operator for the specified Fock space with optional data type `T`.
 """
-function number(b::FockBasis)
-    diag = complex.(b.offset:b.N)
+function number(::Type{T}, b::FockBasis) where T
+    diag = T.(b.offset:b.N)
     data = spdiagm(0 => diag)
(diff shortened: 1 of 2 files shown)


src/fock.jl
```diff
@@ -21,44 +21,63 @@ end
...
-    destroy(b::FockBasis)
+    destroy([T=ComplexF64,] b::FockBasis)
 
-Annihilation operator for the specified Fock space.
+Annihilation operator for the specified Fock space with optional data type `T`.
 """
-function destroy(b::FockBasis)
-    diag = complex.(sqrt.(b.offset+1.:b.N))
+function destroy(::Type{C}, b::FockBasis) where C
+    T = real(C)
```
david-pl, line 43: Could you explain why you're taking the detour over `real(C)` here?

myrddin89: Sorry I was a bit busy today. I always struggle with this issue. One major concern I usually have is to get the right type in computations, especially when using extended precision datatypes (the reason why I have proposed these changes in the first place). 

Assuming that one specifies the datatype as `Complex{T}` for some `T` we still needs to compute the `sqrt`s with the correct precision, as by default `sqrt` uses `Float64` if the input type is not an `AbstractFloat`.

Another path we could follow, is to always assume the specified type is real and construct the state/operator with the correct complex type.

david-pl: Thanks for the explanation, that makes sense.

david-pl:
The changes so far look good to me!

PR #27, opened 2021-05-17
Fix embed method for operators acting on different bases

david-pl:
@amilsted does that fix the issue you were having (mentioned on gitter)?


amilsted:
Hi @david-pl, yes! That fixes the problem. Thank you!

PR #30, opened 2021-07-12
Type-stable constructor for LazyTensor

david-pl:
This is actually a breaking change, since `LazyTensor` now requires the `indices` and `operators` to be sorted before construction. Previously if the arguments were unsorted, it was resorted in the constructor. Problem is, that causes type instabilities. I've never really encountered a case where unsorted arguments were important, so I'm inclined to just break it.

@amilsted you raised this issue in gitter, does that PR solve the inference problem for your use-case? Any other comments?

### DIFF
--- a/src/operators_dense.jl
+++ b/src/operators_dense.jl
@@ -23,6 +23,7 @@ Operator(b::Basis,data) = Operator(b,b,data)
 
 Base.zero(op::Operator) = Operator(op.basis_l,op.basis_r,zero(op.data))
 Base.eltype(op::Operator) = eltype(op.data)
+Base.eltype(::Type{T}) where {BL,BR,D,T<:Operator{BL,BR,D}} = eltype(D)
 Base.size(op::Operator) = size(op.data)
 Base.size(op::Operator, d::Int) = size(op.data, d)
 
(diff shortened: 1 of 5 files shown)


amilsted:
Thanks @david-pl, this is a big improvement! Here is what I get:
```
b1 = FockBasis(3)
b2 = GenericBasis(2)

o1 = DenseOperator(b1, randn(4,4))
o2 = DenseOperator(b2, randn(2,2))

b = b1 ⊗ b2

@code_warntype LazyTensor(b, 1, o1)  # Works!
@code_warntype LazyTensor(b, [1, 2], [o1, o2])  # Fails...
@code_warntype LazyTensor(b, [1, 2], (o1, o2))  # Works!
```

david-pl:
Cool, thanks for checking.

> I guess maybe the operator vector gets an abstract element type, so the converted tuple does too?

Yes, that's true. However, even if you make `b2=FockBasis(1)` which gives the vector `[o1,o2]` a concrete type the inference fails. This is because `LazyTensor` always works with tuples internally so the vector is splat into a tuple. I'm not sure if there is a type-stable way to do this. Maybe we should skip the splatting and just use the vector if you construct with it. This would mean that abstract-typed vectors are kept, which may be undesirable. Another option would be to simply disallow vectors. Opinions?

amilsted:
Disallowing vectors seems okay to me. The user can always splat if they want?

david-pl:
Alright, I deprecated the constructor using vectors, so now you'll see a warning if you don't use tuples.

PR #44, opened 2022-04-30
add ket_l and bra_r for LazyProduct

Lightup1:


src/operators_lazyproduct.jl
```diff
@@ -10,23 +10,34 @@ The factors of the product are stored in the `operators` field. Additionally a
...
+        bra_r=[Bra(operator.basis_r) for operator in operators]
+        new(operators[1].basis_l, operators[end].basis_r, factor, operators,ket_l,bra_r)
     end
 end
 function LazyProduct(operators::T, factor::F=1) where {T,F}
     BL = typeof(operators[1].basis_l)
     BR = typeof(operators[end].basis_r)
-    LazyProduct{BL,BR,F,T}(operators, factor)
+    KTL = typeof([Ket(operator.basis_l) for operator in operators])
+    BTR = typeof([Bra(operator.basis_r) for operator in operators])
```
david-pl, line 34: This will cause the entire list of `Ket`s and `Bra`s to be calculated twice when constructing a `LazyProduct`. Can you do the type inference and construction of the cache in one go?

src/operators_lazyproduct.jl
```diff
@@ -62,28 +73,21 @@ permutesystems(op::LazyProduct, perm::Vector{Int}) = LazyProduct(([permutesystem
...
-    mul!(tmp1,a.operators[end],b,a.factor,0)
+    mul!(a.ket_l[end],a.operators[end],b,a.factor,0)
     for i=length(a.operators)-1:-1:2
-        tmp2 = Ket(a.operators[i].basis_l)
-        mul!(tmp2,a.operators[i],tmp1)
-        tmp1 = tmp2
+        mul!(a.ket_l[i],a.operators[i],a.ket_l[i+1])
     end
-    mul!(result,a.operators[1],tmp1,alpha,beta)
+    mul!(result,a.operators[1],a.ket_l[2],alpha,beta)
```
david-pl, line 81: You're actually never using `ket_l[1]` (as it directly goes into the result). Can you remove it, i.e. never store it at all in the construction?

src/operators_lazyproduct.jl
```diff
@@ -62,28 +73,21 @@ permutesystems(op::LazyProduct, perm::Vector{Int}) = LazyProduct(([permutesystem
...
-    mul!(tmp1,a,b.operators[1],b.factor,0)
+    mul!(b.bra_r[1],a,b.operators[1],b.factor,0)
     for i=2:length(b.operators)-1
-        tmp2 = Bra(b.operators[i].basis_r)
-        mul!(tmp2,tmp1,b.operators[i])
-        tmp1 = tmp2
+        mul!(b.bra_r[i],b.bra_r[i-1],b.operators[i])
     end
-    mul!(result,tmp1,b.operators[end],alpha,beta)
+    mul!(result,b.bra_r[end-1],b.operators[end],alpha,beta)
```
david-pl, line 90: Similar to above, you're not using `bra_r[end]`, so you can actually remove that from the cache.

david-pl: Also, tests fail here if `length(bra_r) == 1` as you're trying to index at `0` then.

src/operators_lazyproduct.jl
david-pl: There is also a test failure here, which happens when `length(ket_l) == 1`, so indexing at `2` errors.

david-pl (changes requested):
@Lightup1 Changes look quite good overall, but I had some comments. Could you adapt your code? Most importantly, there is an issue in `mul!` when the cache only contains one `Ket`/`Bra`. Please fix that.

src/operators_lazyproduct.jl
Lightup1: OK, it can be quickly resovled.

src/operators_lazyproduct.jl
Lightup1: Yes, it seems cause index wrong. I am trying to fix it. And I notice that in the master branch. For one operator `LazyProduct`, it seems `mul!` twice.

src/operators_lazyproduct.jl
```diff
@@ -10,23 +10,34 @@ The factors of the product are stored in the `operators` field. Additionally a
...
-    function LazyProduct{BL,BR,F,T}(operators::T, factor::F=1) where {BL,BR,F,T}
+    ket_l::KTL
+    bra_r::BTR
+    function LazyProduct{BL,BR,F,T,KTL,BTR}(operators::T, factor::F=1) where {BL,BR,F,T,KTL,BTR}
         for i = 2:length(operators)
             check_multiplicable(operators[i-1], operators[i])
         end
-        new(operators[1].basis_l, operators[end].basis_r, factor, operators)
+        ket_l=[Ket(operator.basis_l) for operator in operators]
+        bra_r=[Bra(operator.basis_r) for operator in operators]
```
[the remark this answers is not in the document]
Lightup1: Change to `Tuple`s seems unacceptable, for we need use them in `mul!`.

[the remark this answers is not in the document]
david-pl: I don't understand. The list of `ket_l` is a vector of kets. If there are different bases involved, accessing elements of it will be type-unstable. You can change that vector to a tuple, however, and things should still work fine. All you need is to be able to iterate `ket_l` in `mul!`.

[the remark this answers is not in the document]
Lightup1: OK, I know your point. I'll try it.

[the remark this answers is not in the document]
Lightup1: Sorry, I just find that I can edit the inside vector.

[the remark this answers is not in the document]
Lightup1: Hi @david-pl , I have changed that. 
Thank you very much for your kind instructions.
Learned a lot.

david-pl:
@Lightup1 great, thanks for the quick updates and everything. Things look good to me now.

PR #46, opened 2022-05-14
Turbo complex elmul

Lightup1:
Threaded way to gain speed up using LoopVectorization when mul! FFT operators.

Vector length 2^14

```
BenchmarkTools.Trial: 10000 samples with 1 evaluation.
 Range (min … max):  56.900 μs … 195.600 μs  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     74.100 μs               ┊ GC (median):    0.00%
 Time  (mean ± σ):   73.955 μs ±   6.506 μs  ┊ GC (mean ± σ):  0.00% ± 0.00%

                   ▁▄▅▃▃    ▂▃▅▆█▇▃▁
  ▁▁▁▂▃▂▂▁▁▂▂▂▂▃▃▂▆███████▆▇█████████▇▇▇▇▆▆▅▄▃▃▂▂▂▁▁▁▁▁▁▁▁▁▁▁▁ ▃
  56.9 μs         Histogram: frequency by time         92.8 μs <

 Memory estimate: 0 bytes, allocs estimate: 0.
```

after turbo:
```
BenchmarkTools.Trial: 10000 samples with 1 evaluation.
 Range (min … max):  40.900 μs …  1.389 ms  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     52.800 μs              ┊ GC (median):    0.00%
 Time  (mean ± σ):   58.893 μs ± 40.906 μs  ┊ GC (mean ± σ):  0.00% ± 0.00%

    ▃▆▇█▇▇▆▅▄▃▂▂▁ ▁▁▁▁▁                                       ▂
  █▇██████████████████████▇▇▇▇█▇▆▆▆▅▄▆▆▅▄▄▄▄▅▄▅▄▄▄▃▅▃▄▅▄▄▃▄▃▄ █
  40.9 μs      Histogram: log(frequency) by time       152 μs <

 Memory estimate: 0 bytes, allocs estimate: 0.
```

### DIFF
--- a/Manifest.toml
+++ b/Manifest.toml
@@ -1,0 +1,439 @@
+# This file is machine-generated - editing it directly is not advised
+
+julia_version = "1.7.2"
+manifest_format = "2.0"
+
+[[deps.AbstractFFTs]]
+deps = ["ChainRulesCore", "LinearAlgebra"]
+git-tree-sha1 = "6f1d9bc1c08f9f4a8fa92e3ea3cb50153a1b40d4"
+uuid = "621f4979-c628-5d54-868e-fcf4e3e8185c"
+version = "1.1.0"
+
+[[deps.Adapt]]
+deps = ["LinearAlgebra"]
+git-tree-sha1 = "af92965fb30777147966f58acb05da51c5616b5f"
+uuid = "79e6a3ab-5dfb-504d-930d-738a2a938a0e"
+version = "3.3.3"
+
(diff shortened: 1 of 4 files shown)


Lightup1:
LoopVectorization seems to have more requires on cpu than general packages
Should I close the issue and just use it locally?

PR #51, opened 2022-08-16
Assorted operator fixes

amilsted:
* Add `isequal()` to enable fast comparisons of `===` identical operators
* Fix `==` comparison for LazyTensor so that it checks the subsystem indices
* Better support for LazySum with zero terms (a lazy zero operator)
* Fix `dense(o)` and `sparse(o)` where `o` is a LazySum with a single term.
* Fix LazyProduct multiplication in case there is only one factor
* Make embedding work (at least in some cases) for LazyTensor and LazySum operators


amilsted:
QO tests pass.

david-pl (approved):
@amilsted thanks for the fixes. Everything looks good to me!

PR #53, opened 2022-08-25
Fix for mul!

AmitRotem:
Fix for multiplying LazyTenzor with adjoint dense Operator fails.

note that multipling;
LazyTenzor with dense Operator
adjoint LazyTenzor with dense Operator
adjoint LazyTenzor with adjoint dense Operator
worked ok


amilsted:
Thanks Amit! We should add a test that covers this case to the lazytensor tests file.

PR #56, opened 2022-09-16
fix subspace projector for general state

### LINKED ISSUE #55: projector function of subsbace  should use conjugate transpose
Simple example:
```julia
b = FockBasis(5)
x=Ket(b,[1+1im,1-0.5im,0,0,0,0]/norm([1+1im,1-0.5im,0,0,0,0]))
b_sub = SubspaceBasis(b, [x])

P = projector(b_sub, b)

x_prime = P*x
abs(x_prime.data[1]) # this one should close to one but not
```

`projector` function here:
```julia
function projector(::Type{T}, b1::SubspaceBasis, b2::Basis) where T
    if b1.superbasis != b2
        throw(ArgumentError("Second basis has to be the superbasis of the first one."))
    end
    data = zeros(T, length(b1), length(b2))
    for (i, state) = enumerate(b1.basisstates)
        data[i,:] = state.data # here should use state.data'
    end
    return Operator(b1, b2, data)
end
```

Lightup1:
This should fix #55

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 /docs/build/
+Manifest.toml
--- a/src/subspace.jl
+++ b/src/subspace.jl
@@ -8,8 +8,8 @@ struct SubspaceBasis{S,B,T,H,UT} <: Basis
     superbasis::B
     basisstates::Vector{T}
     basisstates_hash::UT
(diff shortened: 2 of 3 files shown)


Lightup1:
What does `codecov/patch ` mean?
I just used auto-format when I saved files

PR #60, opened 2022-09-30
Faster lazy ops

amilsted:
Reduce compile time of some lazy op arithmetic ops.

For example
```
ls = LazySum(b, b)
op = LazySum(randoperator(b))

@time for _ in 1:N  # compilation time scales with N
    ls += op  # every `+` will require compilation, as the first argument changes type each iteration
end
```

### DIFF
--- a/src/operators_lazysum.jl
+++ b/src/operators_lazysum.jl
@@ -34,12 +34,14 @@ LazySum(::Type{T}, basis_l::Basis, basis_r::Basis) where T = LazySum(basis_l,bas
 LazySum(basis_l::Basis, basis_r::Basis) = LazySum(ComplexF64, basis_l, basis_r)
 
 function LazySum(factors, operators)
+    if operators isa AbstractVector
+        @info "LazySum operators with vector storage of operators may not perform well in time evolution."
+    end
     Tf = promote_type(eltype(factors), mapreduce(eltype, promote_type, operators))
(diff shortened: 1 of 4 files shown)


amilsted:
The main feature of this PR is to avoid getting all component operator types on every `+` call. Instead, we now have a `LazySum` constructor that takes the coefficient type as an argument and we propagate that information when doing arithmetic (`+`, `-`, etc.). With that, the compile time of the example should scale only *linearly* with `N`.

amilsted:
Still would like to find ways to further reduce the time for each addition.

amilsted:
@david-pl I have made it easier to construct LazySum operators with vector storage of operators. Perhaps this is now too easy to do, as code that previously generated tuple LazySums now generate Vector LazySums. Interested in your thoughts.

Vector LazySums behave much better under arithmetic ops (much less compilation), but are typically slower in `mul!()` due to runtime dispatch. I want to be able to build vector LazySums with many terms and later convert them to tuple LazySums for simulations.

david-pl:
@amilsted I've nothing against vector storage here if it has a use-case. So long as the default `LazySum(A,B)` still produces tuples, I'm fine with it.

One question regarding your changes: have you considered using the `BASIS_CHECK` ref value instead of the `skip_checks` keyword? That would allow skipping over checks using the `@samebases` macro when needed.

PR #62, opened 2022-10-10
Add reduced() to compute reduced states by specifying indices to keep.

amilsted:


src/operators_dense.jl
```diff
@@ -192,6 +192,9 @@ function ptrace(psi::Bra, indices)
     return Operator(b_, b_, result)
 end
 
+_index_complement(b::CompositeBasis, indices) = setdiff(1:length(b.bases), indices)
```
david-pl, line 195: You could use `sortedindices.complement(length(b.bases), indices)` here. Not sure if your version is more efficient. If it is, could you replace/update `sortedindices.complement` so we have the best version there?

PR #65, opened 2022-10-14
Operator method to concatenate Kets + tests

AmitRotem:
This is useful to stack Kets before solving ODE with schroedinger_dynamic


AmitRotem:
Any hidden comments?

src/operators_dense.jl
```diff
@@ -20,6 +20,9 @@ end
 Operator{BL,BR}(basis_l::BL,basis_r::BR,data::T) where {BL,BR,T} = Operator{BL,BR,T}(basis_l,basis_r,data)
 Operator(basis_l::BL,basis_r::BR,data::T) where {BL,BR,T} = Operator{BL,BR,T}(basis_l,basis_r,data)
 Operator(b::Basis,data) = Operator(b,b,data)
+Operator(qet::Ket{B}...) where B = Operator(qet[1].basis, GenericBasis(length(qet)), hcat(getfield.(qet,:data)...))
+Operator(basis_r::Basis,qet::Ket{B}...) where B = Operator(qet[1].basis, basis_r, hcat(getfield.(qet,:data)...))
```
Krastanov, line 24: I believe this is raising an "unbound type parameter" error in the tests. Check the failing reports. The easiest way to fix that is probably simply dropping the `where B`, i.e. just use `qet::Ket...`. For more about this you can read https://github.com/JuliaTesting/Aqua.jl/issues/87

AmitRotem: I realize now that a method for Operator with no input at all is a bit weird.
Changed it so it has to get a least one input state.

amilsted:
> Any hidden comments?

Yes, sorry. I started a review and didn't finish it. I will post them now.

src/operators_dense.jl
```diff
@@ -20,6 +20,9 @@ end
 Operator{BL,BR}(basis_l::BL,basis_r::BR,data::T) where {BL,BR,T} = Operator{BL,BR,T}(basis_l,basis_r,data)
 Operator(basis_l::BL,basis_r::BR,data::T) where {BL,BR,T} = Operator{BL,BR,T}(basis_l,basis_r,data)
 Operator(b::Basis,data) = Operator(b,b,data)
+Operator(qet1::Ket, qetva::Ket...) = Operator(qet1.basis, GenericBasis(length(qetva)+1), qet1, qetva...)
+Operator(basis_r::Basis,qet1::Ket,qetva::Ket...) = Operator(qet1.basis, basis_r, qet1, qetva...)
+Operator(basis_l::BL,basis_r::BR,qet1::Ket,qetva::Ket...) where {BL,BR} = Operator{BL,BR}(basis_l, basis_r, hcat(qet1.data, getfield.(qetva,:data)...))
```
amilsted, line 25: While this will work fine, compilation might be very slow for some use cases, such as constructing unitaries from sets of Kets. If the dimension of the space is large, there will be a large number of arguments to this function. Could you add a `Operator(qets::AbstractVector{<:Ket})` or similar and use `reduce(hcat, getfield.(qets, :data)))` internally to avoid the splat?

PR #69, opened 2022-10-21
LazyKet for QuantumCumulants.jl

ChristophHotter:
Definition of LazyKets for the numeric conversion of initial product states in QuantumCumulants.jl.


david-pl:
@Krastanov do you have an idea what the JET CI failures are about here?

Krastanov:
Not sure, they all seem to detect a variety of issues, and it will take some effort to figure out which ones are false positives. It seems most of these also get triggered on master, so presumably with some previous merge we inadvertently added them. It is also possible that with Julia 1.10 now out we simply have more issues being detectable thanks to improvements in static analysis.

I would suggest just bumping the bound of "known issues" in the test file. The fact it is not zero will still be caught by `@test_broken`.

david-pl:
@Krastanov okay, thanks for the quick response :+1:

PR #71, opened 2022-10-23
add JET and Aqua static analysis tests

Krastanov:
Edit: this PR now completely integrates Aqua and JET static analysis in the test framework and fixes various (minor) issues detected by these tools.

~~This PR is not particularly useful on its own, but it makes it easier to start using Aqua and JET for static analysis. JET in particular is a great way to find bugs.~~

- [Aqua](https://github.com/JuliaTesting/Aqua.jl) is just for sanity checks. It did uncover a large number of method ambiguities, but some of them are unimportant. I have not tried to fix them for now and have just marked the test as broken.
- Aqua also complained about exporting symbols that are not actually defined. It seems they were used for documenting. I turned them into comments.
- JET found a ton of issues, but many of them are false positives. JET can be configured to skip these false positives, but that would take some more work. It seems there might be a couple of real issues in there too, but probably not worth investigating before filtering the false positives
- I added JET to a new CI run. It is not turned on by default because JET very frequently breaks on nightly which would lead to QuantumOpticsBase not being tracked by nanosoldier.

This can be merged without negative effects to the code base, but for it to actually be useful some more work needs to be done.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -23,6 +23,7 @@ julia = "1.3"
 [extras]
 Adapt = "79e6a3ab-5dfb-504d-930d-738a2a938a0e"
 Aqua = "4c88cf16-eb10-579e-8560-4a9242c79595"
+Dates = "ade2ca70-3891-5945-98fb-dc099432e06a"
 FFTW = "7a1cc6ca-52ef-59f5-83cd-3a7055c09341"
 JET = "c3a54625-cd67-489e-a8e7-0a5a0ff4e31b"
 LRUCache = "8ac3fa9e-de4c-5943-b1dc-09c6b5f20637"
@@ -34,4 +35,4 @@ Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 UnsafeArrays = "c4a57d5a-5b31-53a6-b365-19f8c011fbd6"
 
 [targets]
-test = ["LinearAlgebra", "SparseArrays", "Random", "Test", "Aqua", "JET", "Adapt", "FFTW", "LRUCache", "Strided", "UnsafeArrays"]
+test = ["LinearAlgebra", "SparseArrays", "Random", "Test", "Aqua", "JET", "Adapt", "Dates", "FFTW", "LRUCache", "Strided", "UnsafeArrays"]
--- a/Project.toml
+++ b/Project.toml
@@ -21,10 +21,17 @@ UnsafeArrays = "1"
 julia = "1.3"
 
 [extras]
+Adapt = "79e6a3ab-5dfb-504d-930d-738a2a938a0e"
+Aqua = "4c88cf16-eb10-579e-8560-4a9242c79595"
+FFTW = "7a1cc6ca-52ef-59f5-83cd-3a7055c09341"
+JET = "c3a54625-cd67-489e-a8e7-0a5a0ff4e31b"
+LRUCache = "8ac3fa9e-de4c-5943-b1dc-09c6b5f20637"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
 SparseArrays = "2f01184e-e22b-5df5-ae63-d93ebab69eaf"
+Strided = "5e0ebb24-38b0-5f93-81fe-25c709ecae67"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
+UnsafeArrays = "c4a57d5a-5b31-53a6-b365-19f8c011fbd6"
 
 [targets]
-test = ["LinearAlgebra", "SparseArrays", "Random", "Test"]
+test = ["LinearAlgebra", "SparseArrays", "Random", "Test", "Aqua", "JET", "Adapt", "FFTW", "LRUCache", "Strided", "UnsafeArrays"]
--- a/src/QuantumOpticsBase.jl
+++ b/src/QuantumOpticsBase.jl
@@ -3,31 +3,46 @@ module QuantumOpticsBase
 using SparseArrays, LinearAlgebra, LRUCache, Strided, UnsafeArrays
 import LinearAlgebra: mul!, rmul!
 
-export bases, Basis, GenericBasis, CompositeBasis, basis,
+export Basis, GenericBasis, CompositeBasis, basis,
         tensor, ⊗, permutesystems, @samebases,
-        states, StateVector, Bra, Ket, basisstate, sparsebasisstate, norm,
+        #states
+                StateVector, Bra, Ket, basisstate, sparsebasisstate, norm,
                 dagger, normalize, normalize!,
-        operators, AbstractOperator, DataOperator, expect, variance,
-            identityoperator, ptrace, reduced, embed, dense, tr, sparse,
-        operators_dense, Operator, DenseOperator, DenseOpType, projector, dm,
-        operators_sparse, SparseOperator, diagonaloperator, SparseOpType,
-        operators_lazysum, LazySum,
-        operators_lazyproduct, LazyProduct,
-        operators_lazytensor, LazyTensor, lazytensor_use_cache, lazytensor_clear_cache,
-        lazytensor_cachesize, lazytensor_disable_cache, lazytensor_enable_cache,
-        superoperators, SuperOperator, DenseSuperOperator, DenseSuperOpType,
+        #operators
+                AbstractOperator, DataOperator, expect, variance,
+                identityoperator, ptrace, reduced, embed, dense, tr, sparse,
+        #operators_dense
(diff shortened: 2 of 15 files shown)


Project.toml
```diff
@@ -21,10 +21,17 @@ UnsafeArrays = "1"
...
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
 SparseArrays = "2f01184e-e22b-5df5-ae63-d93ebab69eaf"
+Strided = "5e0ebb24-38b0-5f93-81fe-25c709ecae67"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
+UnsafeArrays = "c4a57d5a-5b31-53a6-b365-19f8c011fbd6"
 
 [targets]
-test = ["LinearAlgebra", "SparseArrays", "Random", "Test"]
+test = ["LinearAlgebra", "SparseArrays", "Random", "Test", "Aqua", "JET", "Adapt", "FFTW", "LRUCache", "Strided", "UnsafeArrays"]
```
Krastanov, line 37: all these explicit extra dependencies are necessary for JET. They are detected during the static analysis as implicit dependencies but can not be imported for the analysis to complete without them being explicit.

Krastanov:
I believe with the last two commits this is now a valuable linting cleanup of the code base (also including tests that now pass). All of the useful linting is done with Aqua. There are tests marked as broken done with the JET static analysis tool and tests on julia-nightly too, but more work and fixes are needed for them to be enabled.

The "unbound type parameters" commit mostly takes care of signatures that assume `Tuple{Varargs}` that neglect the existance zero-varargs `Tuple{}` (see also  https://github.com/JuliaTesting/Aqua.jl/issues/87 )

The "fix method ambiguities" commit fixes around 20 method ambiguities that used to exist. Anecdotally, I have observed such fixes to have minor positive impact in decreasing compilation times and JET abstract interpretation issues.

I do not think there is anything particularly invasive in this pull request. Could it be considered for merging now?

src/QuantumOpticsBase.jl
```diff
@@ -3,31 +3,46 @@ module QuantumOpticsBase
...
-        operators, AbstractOperator, DataOperator, expect, variance,
-            identityoperator, ptrace, reduced, embed, dense, tr, sparse,
-        operators_dense, Operator, DenseOperator, DenseOpType, projector, dm,
-        operators_sparse, SparseOperator, diagonaloperator, SparseOpType,
-        operators_lazysum, LazySum,
-        operators_lazyproduct, LazyProduct,
-        operators_lazytensor, LazyTensor, lazytensor_use_cache, lazytensor_clear_cache,
-        lazytensor_cachesize, lazytensor_disable_cache, lazytensor_enable_cache,
-        superoperators, SuperOperator, DenseSuperOperator, DenseSuperOpType,
+        #operators
```
Krastanov, line 11: `operators` and other commented out symbols did not actually exist... I assume they were used to comment the code, saying where something is coming from.

src/spinors.jl
```diff
@@ -39,6 +39,7 @@ function directsum(b1::SumBasis, b2::SumBasis)
     bases = [b1.bases...;b2.bases...]
     return SumBasis(shape, (bases...,))
 end
+directsum() = GenericBasis(0)
```
Krastanov, line 42: Similar to the varargs issue: Varargs implies that it can have zero arguments, but the already defined method could not deal with zero arguments.

Krastanov:
And the last three commits now even enables JET and cleans up the warnings detected by it.

src/operators.jl
```diff
@@ -215,13 +215,13 @@ specifies in which subsystems the corresponding operator is defined.
...
     @assert length(basis_l.bases) == length(basis_r.bases)
-    N = length(basis_l.bases)
+    N = length(basis_l.bases)::Int # type assertion to help type inference
     if length(operators) == 0
         return identityoperator(T, basis_l, basis_r)
     end
     indices, operator_list = zip(operators...)
     operator_list = [operator_list...;]
-    indices_flat = [indices...;]
+    indices_flat = [indices...;]::Vector{Int} # type assertion to help type inference
```
Krastanov, line 224: This particular type assertion hint might be unnecessary in the future if improvements to Base.Generator happen. See https://github.com/aviatesk/JET.jl/issues/413

amilsted:
Hi @Krastanov, this is interesting! I haven't used either tool myself yet. Are you finding that JET finds genuine type instabilities and similar issues? The diff suggests that there were some...?

Krastanov:
In the case of QuantumOpticsBase, there were one or two instabilities that were found, but they were mostly inconsequential. Basically, this code base was already good enough, and JET did not find anything too exciting.

Aqua is more close to a linter, not a hardcore static analysis tool like JET. Aqua did fine a tone of method ambiguities that I cleaned up, but that is less of a "bang for the buck".

In other code bases having JET and Aqua has been extremely useful, detecting bugs and helping with refactoring, and keeping the code in a shape that the julia compiler is happier with (faster compilation and better caching).

Krastanov:
If this gets merged, I will do the same treatment to QuantumOptics. It is easier though if the base library is already linted.

FYI, JET and julia are very closely tied and JET depends on Julia internals. That is why I have set JET to test only on nightly (it detects much more), and also why it is not part of the default CI (because JET breaks very often on nightly).

Krastanov:
For some cute bragging rights, inference issues detected in QuantumOptcsBase led to improvements to base Julia https://github.com/JuliaLang/julia/pull/47628

Krastanov:
The was a minor conflict in the export statements due to recently merged changes. That conflict should be fixed now.

Any verdict on this PR?

Krastanov:
The drops in coverage are due to new error-raising methods being defined, i.e. due to lack of some trivial `@test_throw`. I do not know what QuantumOpticsBase policy is on "testing" this type of trivial methods. The tests themselves would simply be copies of the one-line methods, so I avoided writing them on purpose.

david-pl:
I'm all in favor of giving this a shot, it could prove quite useful. Let's keep it for some testing just in QuantumOpticsBase for now, and later we can also include it in QuantumOptics.

I have one question: why do we need to run the JET CI 5 times on different OS (or OS architectures)?

Krastanov:
We do not, that was just an omission on my part. Now JET is executed on one single configuration.

src/operators_lazysum.jl
```diff
@@ -129,7 +129,7 @@ end
 
 identityoperator(::Type{<:LazySum}, ::Type{S}, b1::Basis, b2::Basis) where S<:Number = LazySum(identityoperator(S, b1, b2))
 
-function embed(basis_l::CompositeBasis, basis_r::CompositeBasis, indices, op::LazySum)
+function embed(basis_l::CompositeBasis, basis_r::CompositeBasis, indices::AbstractVector, op::LazySum)
```
amilsted, line 132: I guess this doesn't break any test, but why do we need it?

Krastanov: It leads to a method ambiguity if it is not restricted:

```
1 ambiguities found
Ambiguity #1
embed(basis_l::QuantumOpticsBase.CompositeBasis, basis_r::QuantumOpticsBase.CompositeBasis, index::Integer, op::QuantumOpticsBase.AbstractOperator) @ QuantumOpticsBase ~/Documents/ScratchSpace/quantumoptics/QuantumOpticsBase/src/operators.jl:173
embed(basis_l::QuantumOpticsBase.CompositeBasis, basis_r::QuantumOpticsBase.CompositeBasis, indices, op::QuantumOpticsBase.LazySum) @ QuantumOpticsBase ~/Documents/ScratchSpace/quantumoptics/QuantumOpticsBase/src/operators_lazysum.jl:132
```

There are probably other feasible ways to fix this if it is a concern

amilsted: I think we might be able to kill the first method listed instead.

amilsted: Is it enough to provide `embed(basis_l::CompositeBasis, basis_r::CompositeBasis, indices::Integer, op::LazySum)` and `embed(basis_l::CompositeBasis, basis_r::CompositeBasis, indices, op::LazySum)`?

src/operators.jl
```diff
@@ -191,7 +191,7 @@ function embed(basis_l::CompositeBasis, basis_r::CompositeBasis,
     data[1] = one(Tnum)
     i = N
     while i > 0
-        if i ∈ index
+        if i == index
```
Krastanov, line 194: this seems to have been an old copy-paste "mistake" when the multi-index embed was being turned into a single-index embed fast path. It does not matter because `::Integer ∈ ::Integer` is the same as `::Integer == ::Integer`

Krastanov:
@amilsted , a new embed for LazySum (method ambiguity resolution) is now added as you suggested

amilsted:
LGTM. Will merge after CI.

PR #73, opened 2022-11-24
Factoring out a common interface

Krastanov:
related to https://github.com/qojulia/QuantumOpticsCore.jl/issues/1

Best to keep overall discussion in https://github.com/qojulia/QuantumOpticsCore.jl/issues/1 and postpone discussing this PR until a decision on the above issue is made.


Krastanov:
I think this is ready for a first-pass review. I will comment on a few points in the diffs

src/QuantumOpticsBase.jl
```diff
@@ -3,6 +3,9 @@ module QuantumOpticsBase
 using SparseArrays, LinearAlgebra, LRUCache, Strided, UnsafeArrays
 import LinearAlgebra: mul!, rmul!
 
+import QuantumInterface: dagger, directsum, ⊕, dm, embed, expect, permutesystems,
+        projector, ptrace, reduced, tensor, ⊗
+
```
Krastanov, line 8: This is 50% of the changes. Simply moving the namespace for a few functions that would be of use to other libraries, like `QSymbolics` and `QuantumClifford`. The docstrings are mostly still kept in `QuantumOpticsBase` and of course the vast majority of method definitions are still in here too.

src/bases.jl
```diff
@@ -1,268 +1,3 @@
-"""
```
Krastanov, line 1: This whole file was moved to `QuantumInterface`. Basically everything that has to do with `Basis` is now there. This is the only code that contains logic, not just naming conventions, that was moved.

I also moved the most common bases there, like FockBasis and PauliBasis. I did not move all of them, as some had extra machinery attached to them that I did not want to immediately try to decouple.

src/operators.jl
```diff
@@ -1,19 +1,7 @@
...
-All deriving operator classes have to define the fields
-`basis_l` and `basis_r` defining the left and right side bases.
-
-For fast time evolution also at least the function
-`mul!(result::Ket,op::AbstractOperator,x::Ket,alpha,beta)` should be
-implemented. Many other generic multiplication functions can be defined in
-terms of this function and are provided automatically.
-"""
-abstract type AbstractOperator{BL,BR} end
+import QuantumInterface: AbstractOperator
```
Krastanov, line 4: Some abstract types were also moved, mostly for the benefit of `QSymbolics`.

src/states.jl
```diff
@@ -1,22 +1,13 @@
...
-"""
-Abstract base class for [`Bra`](@ref) and [`Ket`](@ref) states.
-
-The state vector class stores the coefficients of an abstract state
-in respect to a certain basis. These coefficients are stored in the
-`data` field and the basis is defined in the `basis`
-field.
-"""
-abstract type StateVector{B,T} end
+import QuantumInterface: StateVector, AbstractKet, AbstractBra
```
Krastanov, line 3: Two new abstract types were created as well. Now we have `AbstractKet <: StateVector` and `AbstractBra <: StateVector` also for the benefit of `QSymbolics`. Very few changes related to these new abstract types were performed, all in this file

Krastanov (commented):
@david-pl , in-line in this review I left a few comments explaining why various decisions were made

Krastanov:
Of note, the compat for QuantumInterface is set to the current LTS, julia 1.6, while the compat for QuantumOpticsBase is still at 1.3. I personally find it reasonable to push the compat to at least LTS.

Krastanov:
friendly weekly bump on the QuantumInterface PR

having this in a published version would make it easier for me to work on QuantumSavory

Krastanov:
In case it is not clear what the reporting looks like: On passing checks (no increase in invalidations) you see the following in the github action log

image

Project.toml
```diff
@@ -16,6 +17,7 @@ UnsafeArrays = "c4a57d5a-5b31-53a6-b365-19f8c011fbd6"
 Adapt = "1, 2, 3.3"
 FFTW = "1.2"
 LRUCache = "1"
+QuantumInterface = "0.1.0"
```
david-pl, line 20: Why not "0.1"?

david-pl:
@Krastanov changes here look good to me (left one small comment regarding the compat). I'd now:
* merge this PR
* wait for the invalidations CI on the master branch
* publish a new release

Thanks for the work!

Project.toml
Krastanov: I think it means the same. `0.1.x` implies `[0.1.x, 0.2)` and `0.1` implies `[0.1.0, 0.2)` which are the same for x=0. But maybe I misunderstood https://pkgdocs.julialang.org/v1/compatibility/

david-pl: Oh, my mistake. Thanks for clarifying.

david-pl:
@Krastanov CI with JET appears to be failing after the merge. Should I hold off releasing a new version?

Krastanov:
@david-pl , I think this is fine. This CI would be failing relatively often because it is running on nightly and JET and julia nightly frequently get temporarily out of sync (JET uses some deep julia internals). If the error is "outside of tests" it is probably because of a nightly issue. If it is actually inside the tests, then there is something interesting to look for.

If this is too inconsistent, I am happy to remove JET from CI until I find a better way to keep them in sync?

Krastanov:
Also, the reason this is a whole separate CI is so that it is not executed when Julia devs runs PkgEval. That way intermittent noise in the CI does not pollute the ecosystem CI.

david-pl:
@Krastanov thanks, just wanted to clarify this again to be sure. As far as I can tell it's a test failure, but due to a MethodError in `Core.Compiler`: https://github.com/qojulia/QuantumOpticsBase.jl/actions/runs/3665262051/jobs/6196216342#step:21:286

I created a new release so you should be good to go once the PR is merged.

Krastanov:
Indeed, JET is calling various internal undocumented Julia compiler internals in order to do abstract interpretation. The author of JET is probably the main sources of changes to this part of julia, but every so often JET takes a bit to get updated to use the newly changed internals. Maybe it would be reasonable to simply not run it on nightly, however JET improves so fast right now, that I am happy to deal with the intermittent CI failures among my projects. Happy to adjust the CI to the QuantumOptics preferences though!

PR #77, opened 2023-01-30
LazyTensor: Do not assume intermediates should be Vectors

amilsted:
If, for example, we are acting on CuArrays, we don't want intermediate results to be Vector. In these cases, just do not use the cache.

### DIFF
--- a/src/operators_lazytensor.jl
+++ b/src/operators_lazytensor.jl
@@ -261,11 +261,11 @@ function lazytensor_enable_cache(; maxsize::Int = -1, maxrelsize::Real = 0.0)
     return
 end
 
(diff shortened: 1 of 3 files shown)


Krastanov:
I think it is because we are defining `Base` methods with `QuantumInterface` argument types. A bunch of types were moved out from `QuantumOpticsBase` into a common `QuantumInterface` package so that they can be reused by `QuantumSymbolics`, `QuantumClifford`, and `QuantumSavory`.

The reason this was not caught earlier is because Aqua only recently started testing for piracy.

I suggest that we:
- disable piracy checks in Aqua for now
- create a new issue on the tracker so that we keep track of fixing the piracy problems and re-enabling the check in Aqua

Krastanov:
Here is the same issue in QuantumSavory: https://github.com/Krastanov/QuantumSavory.jl/blob/master/test/runtests.jl#L40

`piracy=false` in `Aqua.test_all` should be enough

PR #83, opened 2023-03-16
Check dimensions in sparse gemm and gemv

amilsted:
Basis type checks are not enough in all cases.

### DIFF
--- a/src/sparsematrix.jl
+++ b/src/sparsematrix.jl
@@ -99,6 +99,9 @@ end
 
 
 function gemm!(alpha, M::SparseMatrixCSC, B::AbstractMatrix, beta, result::AbstractMatrix)
+    size(M, 2) == size(B, 1) || throw(ArgumentError("Dimension mismatch."))
+    size(M, 1) == size(result, 1) || throw(ArgumentError("Dimension mismatch."))
+    size(B, 2) == size(result, 2) || throw(ArgumentError("Dimension mismatch."))
     if iszero(beta)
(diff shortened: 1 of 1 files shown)


david-pl (approved):
@amilsted thanks for this! The code looks perfectly fine.
One thing I wonder: should we add the `BASES_CHECK` flag from QuantumInterface here? I.e. if you'd just add `QuantumInterface.BASES_CHECK[] || ...` before each dimension check, a user could skip the checks using the `@samebases` macro. I think that's fair since then the user just guarantees that the bases match, which means the dimensionality should match as well. Although I know that it's probably not entirely correct from a semantics point of view (we're just looking at a pure `gemm` here without any bases). What do you think?

amilsted:
@david-pl I do think that might be a little unexpected. Dense matrix ops will error out independently of `BASES_CHECK`, so it would be inconsistent to skip these checks in the sparse case only. I think probably the overhead here is also tiny.

Alternatively, I guess we could require that basis types include a type parameter for the size or shape.

david-pl:
@amilsted agreed. Let's keep it like this then.

Regarding `NLevelBasis`: there's no specific reason that we don't have the value type parameter there, it's just a bit of a mess. It'd need some clean up, but I'm not sure which way to go overall. If I'm not mistaken, value type parameters can lead to some overhead due to dynamic dispatches and I'm not sure how large those actually are. Anyway, this would be a separate PR.

Thanks!

PR #84, opened 2023-04-01
JET CI fixes

Krastanov:
To my surprise, the merge of JET as a static analyzer in the CI (a few months ago), actually squashed an incorrect debug flag that I was playing with locally. So... JET was not actually running in CI for quite a while. This enables the JET static analyzer once again.

### DIFF
--- a/.github/workflows/ci-jet.yml
+++ b/.github/workflows/ci-jet.yml
@@ -35,7 +35,7 @@ jobs:
       - uses: julia-actions/julia-buildpkg@v1
       - uses: julia-actions/julia-runtest@v1
         env:
-          QUANTUMSAVORY_JET_TEST: true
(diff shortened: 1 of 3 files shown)


Krastanov:
It seems JET is detecting one new issue. I will have to fix it before enabling JET again.

While on the topic of updating the CI, in this same pull request I am adding a commit that updates the various github actions to newer versions and setting up a dependabot that would automatically do this in the future.

Krastanov:
I moved the JET analysis to run on `latest` instead of on `nightly`. Nightly causes a bit too many phantoms of errors. This means that on every new Julia version we might end up with a handful of new errors being detected by the improved static analyzer, but more importantly, it would avoid some false positives that appear on nightly.

This is now ready for review and potential merge.

amilsted:
Looks fine to me!

PR #85, opened 2023-04-03
Mul adjoint and lazy sum

AmitRotem:
Product of an Operator with an `AbstractMatrix` data and an `AbstractOperator` (`LazySum`) fails on;
https://github.com/qojulia/QuantumOpticsBase.jl/blob/2506d0da922999c2a7c0d28c340caf1b769cffe6/src/operators_dense.jl#L97
trying to build an Operator with data that is AbstractMatrix; e.g., `Hermition` or `Adjoint`.

As @amilsted suggested, using `parent` a few times to strip the data to a concrete type before allocating.

Also promoting the element type to account for `Operators` with `Real` data.

### DIFF
--- a/src/operators_dense.jl
+++ b/src/operators_dense.jl
@@ -89,12 +89,12 @@ Base.isapprox(x::DataOperator, y::DataOperator; kwargs...) = false
 *(a::Operator, b::Number) = Operator(a.basis_l, a.basis_r, b*a.data)
 *(a::Number, b::Operator) = Operator(b.basis_l, b.basis_r, a*b.data)
 function *(op1::AbstractOperator{B1,B2}, op2::Operator{B2,B3,T}) where {B1,B2,B3,T}
-    result = Operator{B1,B3,T}(op1.basis_l, op2.basis_r, similar(op2.data,length(op1.basis_l),length(op2.basis_r)))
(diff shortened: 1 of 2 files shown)


src/operators_dense.jl
```diff
@@ -108,7 +108,7 @@ function *(psi::Bra{BL,T}, op::AbstractOperator{BL,BR}) where {BL,BR,T}
     mul!(result,psi,op)
     return result
 end
-
+_parent = parent∘parent∘parent
```
amilsted, line 111: It occurs to me we could do `_parent(x) = (x === parent(x) ? x : _parent(parent(x)))`. I think maybe this is what you were getting at when we discussed?

amilsted: PS: If you make this change, please rebase on the master branch to get the latest CI config.

amilsted: I am not sure this will be type-stable. May be worth checking. Probably could replace `x === parent(x)` with `typeof(x) == typeof(parent(x))`.

amilsted: Maybe you meant 
```julia
_parent(x::T, x_parent::T) = x
_parent(x, x_parent) = _parent(x_parent, parent(x_parent))
_parent(x) = _parent(x, parent(x))
```
which ought to be type stable :)

AmitRotem: Kinda, but your implementation has the added benefit of actually working!

david-pl: Nice, I saw the original version of this in the PR and thought about a better solution. Came up with the current version too.

PR #86, opened 2023-04-07
Request For Comment: Arithemtic between Operators and LazyOperators

mabuni1998:
Extend some of the current operation with LazyOperators to allow for smoother interface with e.g. LazyTensor. As an example I defined +(a::LazyTensor,LazyTensor) so that it returns LazySum. 

Note that tensor(a::Operator,b::LazyTensor) only behaves well if Operator does not contain a CompositeBasis. This could probably be fixed by specifying the type more explicitly and returning error if it does contain CompositeBasis. Other solutions are welcome.


Krastanov:
For context, Matias is making this pull request because he is developing an extension to QuantumOptics that needs it on https://github.com/mabuni1998/WaveguideQED.jl/

Krastanov:
Matias, let's include this in the library itself so that we can see its effect on the test runs. Could you rename it to something that fits the naming conventions of the rest of the files (lower snakecase) and include it here (under the other lazy ops) https://github.com/qojulia/QuantumOpticsBase.jl/blob/master/src/QuantumOpticsBase.jl#L68

After the static analysis from JET and Aqua is done complaining we can start looking into also adding tests.

src/LazyOperatorAritmetic.jl
```diff
@@ -0,0 +1,107 @@
+
+function QuantumOpticsBase.:+(a::LazyTensor{B1,B2},b::LazyTensor{B1,B2}) where {B1,B2}
+    LazySum(a,b)
+end
+
+function QuantumOpticsBase.:-(a::LazyTensor{B1,B2},b::LazyTensor{B1,B2}) where {B1,B2}
+    LazySum([1,-1],[a,b])
```
Krastanov, line 7: Does this still work if you use tuples instead of lists? Asking about it because tuples can be made non-allocating, while Julia can not do that to lists yet. If it is impossible, we should probably file a bug with QuantumOpticsBase

src/LazyOperatorAritmetic.jl
```diff
@@ -0,0 +1,107 @@
...
+        LazyTensor(a.basis_l,a.basis_r,[1],(a,),1) ⊗ b
+    end
+end
+
+
+
+function QuantumOpticsBase.:⊗(a::Operator,b::LazySum)
+    btotal_l = a.basis_l ⊗ b.basis_l
+    btotal_r = a.basis_r ⊗ b.basis_r
+    ops = Vector{AbstractOperator}(undef,length(b.operators))
```
Krastanov, line 58: How feasible is it to (1) use a tuple instead of a Vector and (2) use a type more specific than AbstractOperator?

Krastanov (commented):
I added some initial comments in this review.

src/LazyOperatorAritmetic.jl
mabuni1998: It is indeed possible to use Tuples instead. I have now changed this in a new commit.

src/LazyOperatorAritmetic.jl
mabuni1998: Although not perfect, we now use a Tuple:

https://github.com/mabuni1998/QuantumOpticsBase.jl/blob/f4b872eaaa7b014afa9ad6d502ac81ca756dd805/src/operators_lazyaritmetic.jl#L68-L73

There might be better solutions, but in this we at least use a Tuple.

mabuni1998:
The changes are now included in the library under the file: operators_lazyaritmetic.jl

mabuni1998:
Tests are failing because +(op1::LazyProduct,op2::LazyProduct) no longer throws and error and thus the following line in the test fails:

https://github.com/qojulia/QuantumOpticsBase.jl/blob/b7684f71458bffa26feebab16cd97947fa3702ad/test/test_operators_lazyproduct.jl#L66

I will work on changing this as well as adding a couple of tests ensuring that the arithmetic is working as it should.

Krastanov:
We will probably have a few rounds of test fixing. Looking at the first test that is failing, it seems you are currently a bit permissive with what operators can be summed up. See this now-failing test https://github.com/qojulia/QuantumOpticsBase.jl/blob/master/test/test_operators_lazyproduct.jl#L66 : in it op1 and op2 have different bases so probably this should raise an error.

It would probably be easiest to run the tests locally a couple of times, as the CI tests stop at the first failure, while the default local settings are to run all tests and show you all failures.

src/operators_lazyproduct.jl
```diff
@@ -74,6 +94,16 @@ permutesystems(op::LazyProduct, perm::Vector{Int}) = LazyProduct(([permutesystem
 identityoperator(::Type{LazyProduct}, ::Type{S}, b1::Basis, b2::Basis) where S<:Number = LazyProduct(identityoperator(S, b1, b2))
 
 
+#Assume same basis
```
Krastanov, line 97: What does "assume same basis" means here? Is it something that would lead to bugs if not followed? Can we have tests for it, even if the test is basically just checking that an error is thrown instead of silently continuing with the operation?

test/test_abstractdata.jl
```diff
@@ -412,9 +412,9 @@ x2 = Ket(b_r, rand(ComplexF64, length(b_r)))
 xbra1 = Bra(b_l, rand(ComplexF64, length(b_l)))
 xbra2 = Bra(b_l, rand(ComplexF64, length(b_l)))
 
-# Addition
-@test_throws ArgumentError op1 + op2
-@test_throws ArgumentError op1 - op2
+# Addition Addition of LazyTensor now returns LazySum
```
Krastanov, line 415: Before merge I would suggest deleting the commented-out tests. They would not be informative to future readers, long after this merge. Future readers should use `git blame` if they want to know why a change was made.

Could you add tests for the failure modes of this contribution? E.g., `@test_throws` for where your code specifically raises an error to warn about something it can not do. And are there any other implicit assumptions about shapes and bases and other things? We do not want the user to obtain nonsense results if they do not happen to know that some combination of operations is not supported.

src/QuantumOpticsBase.jl
```diff
@@ -80,5 +80,6 @@ include("metrics.jl")
 include("spinors.jl")
 include("phasespace.jl")
 include("printing.jl")
+#include("operators_lazyaritmetic.jl")
```
Krastanov, line 83: this probably should be deleted before merging

amilsted:
Thanks for this @mabuni1998. I am definitely in favor of making laziness "contagious" in the way you have done here. What do you think @david-pl?

src/operators_lazytensor.jl
```diff
@@ -96,22 +96,48 @@ isequal(x::LazyTensor, y::LazyTensor) = samebases(x,y) && isequal(x.indices, y.i
 # Arithmetic operations
 -(a::LazyTensor) = LazyTensor(a, -a.factor)
 
-function +(a::LazyTensor{B1,B2}, b::LazyTensor{B1,B2}) where {B1,B2}
-    if length(a.indices) == 1 && a.indices == b.indices
-        op = a.operators[1] * a.factor + b.operators[1] * b.factor
-        return LazyTensor(a.basis_l, a.basis_r, a.indices, (op,))
```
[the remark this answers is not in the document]
mabuni1998: My experience was that the previous implementation was very limiting. Especially since the custom operators I have been playing around with were not DataOperators but AbstractOperators, where the operation was defined via a function rather than a matrix. Therefore, these cannot be trivially added (except by using LazySum), and the above implementation fails. Also, length(a.indices) ==1 is required, and I could imagine situations where one would like to be able to add LazyTensors containing more than one operator. 

However, one could perhaps keep the original behavior by dispatching on LazyTensors containing only one DataOperator. That is adding a function like this (draft, I'm not entirely sure it works):

```julia
const single_dataoperator{B1,B2} = LazyTensor{B1,B2,ComplexF64,Vector{Int64},Tuple{T}} where {B1,B2,T<:DataOperator} 
function +(a::T1,b::T2) where {T1 <: single_dataoperator{B1,B2},T2 <: single_dataoperator{B1,B2}}
    if length(a.indices) == 1 && a.indices == b.indices
        op = a.operators[1] * a.factor + b.operators[1] * b.factor
        return LazyTensor(a.basis_l, a.basis_r, a.indices, (op,))
    end
    throw(ArgumentError("Addition of LazyTensor operators is only defined in case both operators act nontrivially on the same, single tensor factor."))
end
```

src/operators_lazyproduct.jl
mabuni1998: What I meant is that BL and BR are the same since I'm not sure the methods defined would work if one considered operators with BL != BR. This is already handled by the dispatch, and the user should just get an error saying that addition for these types of operators is not defined. I will look into this and create specific tests for it.

src/QuantumOpticsBase.jl
mabuni1998: Oh yeah... Done that now ;)

src/operators_lazytensor.jl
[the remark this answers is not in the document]
david-pl: @mabuni1998 I think it's worth trying to keep the original intact as you suggest. If we can handle it via dispatch, we won't lose anything. Or am I missing some case here?

src/operators_lazytensor.jl
```diff
@@ -96,22 +96,48 @@ isequal(x::LazyTensor, y::LazyTensor) = samebases(x,y) && isequal(x.indices, y.i
...
-    end
-    throw(ArgumentError("Addition of LazyTensor operators is only defined in case both operators act nontrivially on the same, single tensor factor."))
+function +(a::LazyTensor{B1,B2},b::LazyTensor{B1,B2}) where {B1,B2}
```
david-pl, line 99: More a generic comment and not strictly necessary in this PR as it works the same: seeing as these methods are quite similar to the `+` methods above, we could (and should) probably create a common abstract type, e.g. `abstract type LazyOperator{BL,BR} <: AbstractOperator{BL,BR} end` and make `LazySum, LazyProduct, LazyTensor <: LazyOperator`. Then we could just define e.g.
```julia
+(a::Operator, b::LazyOperator) = LazySum(a, b)
```
instead of duplicating methods.

Not sure if this will "just work", though, so maybe leave it for another time.

mabuni1998: I completely agree and have thought about defining that before. I can give it a try and see if it "just works".

amilsted: Could alternatively define `LazyOperator = Union{LazySum,LazyProduct,LazyTensor}`.

test/test_abstractdata.jl
mabuni1998: Updated code to no longer assume anything about BL == BR and added test to ensure that errors are thrown if addition between operators is not allowed.

src/operators_lazyproduct.jl
mabuni1998: BL == BR is no longer assumed and code is updated accordingly

src/operators_lazytensor.jl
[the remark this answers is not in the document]
mabuni1998: No I don't think we will lose anything. I have implemented to above as:

https://github.com/mabuni1998/QuantumOpticsBase.jl/blob/64da84f45beb5fc813b4cc5e4afb5ed7a6bf0fc8/src/operators_lazytensor.jl#L99-L113

src/operators_lazytensor.jl
mabuni1998: I ended up choosing LazyOperator as an abstract type. We can change this to LazyOperator = Union{LazySum,LazyProduct,LazyTensor} of course if we want. For now see the following lines where the addition between LazyOperators is defined:

https://github.com/mabuni1998/QuantumOpticsBase.jl/blob/64da84f45beb5fc813b4cc5e4afb5ed7a6bf0fc8/src/operators_lazysum.jl#L82-L104

mabuni1998:
I don't know if this was intended but previously I found that:

```julia
b1a = GenericBasis(2)
b1b = GenericBasis(3)
b2a = GenericBasis(1)
b2b = GenericBasis(4)
b3a = GenericBasis(6)
b3b = GenericBasis(5)
I1 = dense(identityoperator(b1a, b1b))
I2 = dense(identityoperator(b2a, b2b))
I3 = dense(identityoperator(b3a, b3b))

a = tensor(I1,I3)
a == identityoperator(a)
```

would not result in false. To fix it, I implemented identityoperator() when having CompositeBasis. See the following for the implementation:

https://github.com/mabuni1998/QuantumOpticsBase.jl/blob/64da84f45beb5fc813b4cc5e4afb5ed7a6bf0fc8/src/operators_dense.jl#L237-L238

src/operators_dense.jl
```diff
@@ -234,6 +234,9 @@ permutesystems(a::AdjointOperator{B1,B2}, perm) where {B1<:CompositeBasis,B2<:Co
 
 identityoperator(::Type{S}, ::Type{T}, b1::Basis, b2::Basis) where {S<:DenseOpType,T<:Number} =
     Operator(b1, b2, Matrix{T}(I, length(b1), length(b2)))
+identityoperator(::Type{S}, ::Type{T}, b1::CompositeBasis, b2::CompositeBasis) where {S<:DenseOpType,T<:Number} =
+    Operator(b1, b2, kron([Matrix{T}(I, length(b1.bases[i]), length(b2.bases[i])) for i in reverse(1:length(b1.bases))]...))
+
```
amilsted, line 239: This is strange... I am probably missing something obvious, but it looks like these two variants should give exactly equal results for `CompositeBasis`?

mabuni1998: Yes, I agree they should, but from my testing they don't. Try running the following example:

> I don't know if this was intended but previously I found that:
> 
> ```julia
> b1a = GenericBasis(2)
> b1b = GenericBasis(3)
> b2a = GenericBasis(1)
> b2b = GenericBasis(4)
> b3a = GenericBasis(6)
> b3b = GenericBasis(5)
> I1 = dense(identityoperator(b1a, b1b))
> I2 = dense(identityoperator(b2a, b2b))
> I3 = dense(identityoperator(b3a, b3b))
> 
> a = tensor(I1,I3)
> a == identityoperator(a)
> ```
> 
> would result in false. To fix it, I implemented identityoperator() when having CompositeBasis. See the following for the implementation:
> 
> https://github.com/mabuni1998/QuantumOpticsBase.jl/blob/64da84f45beb5fc813b4cc5e4afb5ed7a6bf0fc8/src/operators_dense.jl#L237-L238
> 
> This was important because in LazyTensor we check whether the operator is identityoperator and when the above didn't return true it lead to unwanted errors.
> 
> https://github.com/mabuni1998/QuantumOpticsBase.jl/blob/64da84f45beb5fc813b4cc5e4afb5ed7a6bf0fc8/src/operators_lazytensor.jl#L115-L136

src/operators_lazytensor.jl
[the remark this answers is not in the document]
mabuni1998: Probably won't be performance-critical no, as you are most likely creating the operators once at the beginning of the simulation and then not changing them as you do multiplications etc.

src/operators_dense.jl
amilsted: That's odd. Could it be that the eltype is different? Maybe that `identityoperator()` method is broken?

mabuni1998: It's the data inside; they do not represent the same matrix. Maybe Julias UnitScaling follows another convention than we expect, or maybe the example I'm making shouldn't return true?

To make it even clearer, consider:

```julia
b1a = GenericBasis(2)
b1b = GenericBasis(3)
I1 = identityoperator(b1a, b1b)
I2 = identityoperator(b1b, b1a)
a = tensor(I1,I2)
julia> a.data
6×6 SparseArrays.SparseMatrixCSC{ComplexF64, Int64} with 4 stored entries:
 1.0+0.0im      ⋅          ⋅          ⋅          ⋅          ⋅
     ⋅      1.0+0.0im      ⋅          ⋅          ⋅          ⋅
     ⋅          ⋅          ⋅      1.0+0.0im      ⋅          ⋅
     ⋅          ⋅          ⋅          ⋅      1.0+0.0im      ⋅
     ⋅          ⋅          ⋅          ⋅          ⋅          ⋅
     ⋅          ⋅          ⋅          ⋅          ⋅          ⋅
julia> identityoperator(a).data
6×6 SparseArrays.SparseMatrixCSC{ComplexF64, Int64} with 6 stored entries:
 1.0+0.0im      ⋅          ⋅          ⋅          ⋅          ⋅
     ⋅      1.0+0.0im      ⋅          ⋅          ⋅          ⋅
     ⋅          ⋅      1.0+0.0im      ⋅          ⋅          ⋅
     ⋅          ⋅          ⋅      1.0+0.0im      ⋅          ⋅
     ⋅          ⋅          ⋅          ⋅      1.0+0.0im      ⋅
     ⋅          ⋅          ⋅          ⋅          ⋅      1.0+0.0im
```

amilsted: Oh... Yeah, I think in this case they actually *should not* be equal. It's kind of funny to even define `identityoperator()` in the rectangular case. Clearly it's not the identity!

amilsted: To summarize, the two methods of `identityoperator()` above are only equivalent in case the left and right composite bases both have the same length and have the same number of factors, and the dimensions of the factors match. Otherwise the first definition (the original one) is correct - it actually produces an identity operator int he square case - where the second is not. The second version also fails in case the left basis has more factors than the right.

Should we consider raising a warning if `identityoperator()` is used in rectangular cases?

Krastanov: +1 for a warning

src/operators_lazytensor.jl
```diff
@@ -96,21 +96,45 @@ isequal(x::LazyTensor, y::LazyTensor) = samebases(x,y) && isequal(x.indices, y.i
...
+function tensor(a::LazyTensor{B1,B2},b::AbstractOperator{B3,B4}) where {B1,B2,B3,B4}
+    if isequal(b,identityoperator(b))
+        btotal_l = a.basis_l ⊗ b.basis_l
+        btotal_r = a.basis_r ⊗ b.basis_r
+        LazyTensor(btotal_l,btotal_r,a.indices,(a.operators...,),a.factor)
+    elseif B3 <: CompositeBasis || B4 <: CompositeBasis
+        throw(ArgumentError("tensor(a::LazyTensor{B1,B2},b::AbstractOperator{B3,B4}) is not implemented for B3 or B4 being CompositeBasis unless b is identityoperator "))
+    else
+        a ⊗ LazyTensor(b.basis_l,b.basis_r,[1],(b,),1)
+    end
```
mabuni1998, line 124: How should we treat LazyTensor with non-equivalent LR and BR. The first if statement is only well defined for BL==BR it seems.

src/operators_dense.jl
mabuni1998: I see good point. But then I would like some input on how we should treat LazyTensor with non-equivalent left and right basises. See below

src/operators_lazytensor.jl
amilsted: Yeah, so LazyTensor with non-rectangular factors in the composite basis implicitly defines simple isometries, like you constructed in your new implementation of `identityoperator()`. Honestly I think it's okay to just not handle the `b is an identity` case specially at all and just include the explicit identity/isometry operator. It's kind of unfortunate to explicitly check that a matrix is the identity here anyway if I'm expecting a lazy operation.

I have an `embed_lazy()` that I use to achieve this kind of thing.

Perhaps a nicer solution is to introduce a proper LazyIdentity/LazyIsometry operator that gets materialized only when necessary. Then we could implement `tensor(a::LazyTensor, b::LazyIdentity)` using something like your first conditional branch here.

Krastanov: @amilsted , could you put up a gist with `embed_lazy` or just an incomplete pull request?

This discussion and `embed_lazy` would probably have a bearing on this issue as well: https://github.com/qojulia/QuantumOptics.jl/issues/322

amilsted: https://github.com/qojulia/QuantumOpticsBase.jl/pull/88

src/operators_dense.jl
david-pl: Okay, so that method here should be removed again.

src/operators_lazytensor.jl
david-pl: I'd also say we can ignore the special `isequal(b,identityoperator(b))`. It should (hopefully) be the exception to run into this, so the impact shouldn't be too large.

One alternative way of solving this would be doing something like #90, but that's just an idea. For now I'd say don't special handle it, remove the new `identityoperator` method above and we're good to go.

By the way, thanks for all the work @mabuni1998 and thanks for the reviews @amilsted @Krastanov. I appreciate it!

mabuni1998: You're welcome! I now pushed a new version with the above things fixed. 

I would like to say that originally I put the check for identityoperator there because, in my own code, I noticed performance drops when doing stuff like the following (WaveguideBasis is a custom basis with operators defined as functions, not as matrices, and they, therefore, have to be Lazy): 
```
bc = FockBasis(2)
bw = WaveguideBasis(2,times)
adw = tensor(create(bc),identityoperator(bc),identityoperator(bc),destroy(bw))
```
Because here, identityoperator(bc) is explicitly stored in the resulting LazyTensor (there is a custom method creating LazyTensor here). Probably this is better solved by allowing operators spanning multiple basises to be stored in LazyTensor. Right now, we explicitly check for this case and don't allow it. However, this would require extra work (and I don't know how much) in the mul! method so let's leave it. This problem (I think) would not be solved by implementing LazyIdentity since identityoperator(bc) would not return such abjoct.

Also, I'm aware that in the above, one could just use embed or explicitly define LazyTensor. I simply just wanted to make it easier for the user to have better performance. 
Furthermore, my use case is probably niece, since I'm only using Lazy operators because my operators are defined as functions, whereas in most cases, you use it because evaluating the tensor product is expensive. I can just define custom methods that take care of this.

mabuni1998:
@amilsted @Krastanov @david-pl  Are we ready to merge this?

amilsted:
> @amilsted @Krastanov @david-pl Are we ready to merge this?

I think I'd like to do one more patch release first.

david-pl:
@mabuni1998 @amilsted It's good to merge from my side. @amilsted please merge (or let me know) once you're done with the patch release.

PR #88, opened 2023-04-14
Draft: Embedding with lazy semantics

amilsted:
Basic version.

### DIFF
--- a/src/operators_lazytensor.jl
+++ b/src/operators_lazytensor.jl
@@ -203,7 +203,7 @@ end
 identityoperator(::Type{<:LazyTensor}, ::Type{T}, b1::Basis, b2::Basis) where {T<:Number} = LazyTensor(b1, b2, Int[], Tuple{}(), one(T))
 
 """
-    embed_lazy(b::QOB.Basis, i, op::QOB.AbstractOperator)
+    embed_lazy(b::Basis, i, op::AbstractOperator)
 
 Embed an operator in a larger hilbert space, at site(s) `i` of basis `b`, using a
 lazy representation of the tensor product and preserving LazySum.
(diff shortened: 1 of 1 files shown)


Krastanov:
I am in favor of this getting merged. This new functionality is not exported, so changes to it will still be possible. QuantumSymbolics and QuantumSavory would like to use it already, so there will be feedback on it as well.

amilsted:
Ok, perhaps I'll add some simple tests first.

Krastanov:
Any chance we can remove the `b::Basis` argument? I would like to be able to make an object like "apply CNOT gate to subsystems 2 and 4" without having to commit to how many qubits I have.

amilsted:
> Any chance we can remove the `b::Basis` argument? I would like to be able to make an object like "apply CNOT gate to subsystems 2 and 4" without having to commit to how many qubits I have.

Huh, that's an interesting suggestion. Currently we define operators wrt to a concrete composite basis with integer indices. You could lazily embed operators on two qubits, then lazily embed the result into a larger system later. Not sure if that is currently supported, but it would be easy enough to do for the LazyTensor case.

Another useful feature for this kind of use-case would be a "labeled basis". If we had that, we could map from smaller <-> larger composite systems by label.

Krastanov:
Maybe a first step would be to permit `nothing` or `PlaceholderBasis() <: AbstractBasis` in the basis field and just see how many things break. Then whenever a multiplication or similar operation is performed we create a new lazy object that steals the basis from whatever it is being multiplied with. It is not the most efficient way to do it, but we can do it in a relatively compiler-friendly way (and most importantly, without modifying much of the rest of the library).

Named bases would also be nice, but I consider them separate from this.

Krastanov:
I converted this to draft, just to make it easier to track what I need to review on my github dashboard.

PR #89, opened 2023-04-14
Julia >=1.8 required since v0.3.10

matiot:
QuantumOpticsBase v0.3.10 makes use of the LazyString type, that has been introduced in Julia v1.8. For example, precompiling QuantumOpticsBase v0.3.10 on Julia v1.6.2 returns errors. This change increases the minimal Julia compatible version to v1.8.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -20,7 +20,7 @@ LRUCache = "1"
 QuantumInterface = "0.1.0"
 Strided = "1"
 UnsafeArrays = "1"
-julia = "1.3"
+julia = "1.8"
 
 [extras]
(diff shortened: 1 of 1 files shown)


amilsted:
We could probably avoid using lazy strings instead of doing this.

david-pl:
@matiot I actually agree with @amilsted here. There's only 4 occurrences and they all were introduced with this commit: https://github.com/qojulia/QuantumOpticsBase.jl/commit/2506d0da922999c2a7c0d28c340caf1b769cffe6
FYI, @AmitRotem, do you see any reason why these are absolutely necessary?

AmitRotem:
No, just to make error message more informative

amilsted:
Addressing in #91

Issue #90, opened 2023-04-15
Implement a lazy identity operator


david-pl:
The `identityoperator` function returns a sparse array. As the action of the identity matrix on states and operators is known, however, we could make this lazy. This may not appear to be a large change, but it could potentially have a large impact on the implementation of `LazyTensor` and its `mul!` methods.

I'm not set on actually doing is, it's just an idea that I thought worth discussing.

We could, for example, use [FillArrays](https://github.com/JuliaArrays/FillArrays.jl) to implement this. They export an `Eye` function which lazily represents a matrix that is zero except on the diagonal with corresponding `getindex` methods and such. We could just use that as `.data` field when constructing `identityoperator`.
```julia
using FillArrays
id = Eye{ComplexF64}(4,3)
id isa AbstractMatrix  # true
```
Alternatively, if we don't want to add this dependency, we could also implement it ourselves (it's not super complicated), which would also give us more control over e.g. the `mul!` implementations and such. I'd prefer not duplicating work, however.

The big advantage I see in implementing this is that it might greatly simplify the implementation of `LazyTensor`. If identity operators are efficient and lazy, then we could remove the entire concept of `indices` from it. This will also simplify the underlying `mul!` methods.

The disadvantage I is that this may have a negative impact on `LazyTensor` performance. Without further dispatches on the `Eye` matrix above, for example, we'd be doing a ton of unnecessary multiplication operations. Ultimately the question whether this impacts performance negatively can only be answered if we do the implementation first, so I'm not sure if we want to invest in this.

One option I see for now is to just change the implementation of `identityoperator` to return a lazy `Eye` (or add a new method such as `lazy_identityoperator`) and leave the `LazyTensor` implementation untouched.

david-pl:
Actually, it's not super hard to get a first comparison for performance. If you just throw "lazy" identities in a `LazyTensor` you can see an approximate impact without any further optimizations. Consider the following:

```julia
using QuantumOpticsBase
using FillArrays
using BenchmarkTools

dims = [2,3,5,4]
bs = GenericBasis.(dims)
bcomp = ⊗(bs...)

idx = 3
a = randoperator(bs[idx])

function lazy_identity(b::Basis)
    n = length(b)
    return Operator(b,b,Eye{ComplexF64}(n))
end

A = LazyTensor(bcomp, (idx,), (a,))
op_list = [i == idx ? a : lazy_identity(b) for (i,b) in enumerate(bs)]
[...]

PR #94, opened 2023-04-19
Lazy identities via FillArrays

amilsted:
Makes `identityoperator()` return a `DataOperator` containing `FillArrays.Eye`, which is a lazy representation of an identity matrix, as suggested in #90.

* Adds support for skipping over square identity matrices in `LazyTensor` `mul!()`.
* Replaces `SimpleIsometry`, used internally by LazyTensor `mul!()`, with `Eye`.
* Fixes handling of `beta=0` in `mul!()` on lazy operators.
* Does less work in lazy operator `mul!()` when `alpha==0`.

NOTE: Currently, a tensor product of lazy identities gives us a Diagonal DataOperator, rather than a big Eye.


amilsted:
https://github.com/qojulia/QuantumOptics.jl/blob/master/test/test_timeevolution_schroedinger.jl#L103 now fails because `u0` is a lazy identity. Arguably, one should always use dense operators for propagators?

We could catch calls to `schroedinger()` where the initial state is the wrong type and return a more sensible error.

Krastanov:
> We could catch calls to `schroedinger()` where the initial state is the wrong type and return a more sensible error.

Maybe we can add `schroedinger(tspan, u0::NotDenseOrSomething, ...) = schroedinger(tspan, dense(u0), ...)`

amilsted:
> > We could catch calls to `schroedinger()` where the initial state is the wrong type and return a more sensible error.
> 
> Maybe we can add `schroedinger(tspan, u0::NotDenseOrSomething, ...) = schroedinger(tspan, dense(u0), ...)`

Yeah, we could. I'm slightly wary because apparently sparse operators worked before. I suppose there could be cases for which one knows some symmetry is preserved and so sparsity makes sense..? Simplest example is a diagonal Hamiltonian and initial state.

amilsted:
@Krastanov Do you know the best way to avoid Aqua complaining about [this](https://github.com/JuliaArrays/FillArrays.jl/issues/105#issuecomment-1518406018) without disabling too may tests?

Krastanov:
> @Krastanov Do you know the best way to avoid Aqua complaining about [this](https://github.com/JuliaArrays/FillArrays.jl/issues/105#issuecomment-1518406018) without disabling too may tests?

I will look into this tonight (Apr 22nd)

Krastanov:
> > @Krastanov Do you know the best way to avoid Aqua complaining about [this](https://github.com/JuliaArrays/FillArrays.jl/issues/105#issuecomment-1518406018) without disabling too may tests?
> 
> I will look into this tonight (Apr 22nd)

Actually, it seems you already have set the `exclude` filter. I believe this is the proper way to do it. Although I would suggest excluding `FillArrays` the package, not `Base.reshape` the function.

If not all flags are supported, I occasionally just do `ambiguities=false` and just call the dedicated `test_ambiguities` with more detailed arguments. https://juliatesting.github.io/Aqua.jl/dev/#Aqua.test_ambiguities-Tuple{Any}

david-pl:
@amilsted the code here looks good as far as I can tell. Performance also lives up to the previous one (except for the explicit isometries case), though I only ran some quick checks.

Regarding the test failure in `schroedinger`: Would it be fair to leave the conversion to an explicit sparse/dense operator to the user here? After all, essentially the same error would occur if you attempted in-place multiplication on a `Diagonal` or solving an `ODEProblem` with one as initial state. The case we're seeing is really just a test and I wouldn't expect a lot of cases where people throw an `identityoperator` into a time evolution. We could then just fix the failure downstream. What do you think?

src/operators_lazytensor.jl
```diff
@@ -331,8 +331,13 @@ function _tp_matmul_mid!(result, a::AbstractMatrix, loc::Integer, b, α::Number,
         size(b, loc) == size(a, 2) && size(result, loc) == size(a, 1) || throw(DimensionMismatch("Dimensions of Eye matrix do not match subspace dimensions."))
         d = min(size(a)...)
 
-        rmul!(result, β)
-        @strided result_r[:, 1:d, :] .+= α .* br[:, 1:d, :]
+        if !iszero(β)
```
david-pl, line 334: @amilsted oh no, I completely missed that as well. Looks like we're missing a case in the unit tests too, right?

amilsted: It was in the previous code, I just missed its importance and dropped it while putting it in this PR. Tests did actually catch the problem, but it might be better to have more explicit tests.

amilsted:
Breakage should no longer fail after https://github.com/qojulia/QuantumOptics.jl/pull/362 is merged.

amilsted:
@david-pl @Krastanov Do you think this is okay to merge?

Krastanov:
Looks good to me! Should we make an issue for the note you have?

> Currently, a tensor product of lazy identities gives us a Diagonal DataOperator, rather than a big Eye.

PR #95, opened 2023-04-21
Fix for multiplication of LazySum with no elements

AmitRotem:
Changed `eltype` for `LazySum` so it won't fail on an empty operator due to reducing over empty list.

Also added dim check for multiplication with empty `LazySum`.
Such that `LazySum(NLevelBasis(1),NLevelBasis(1)) * Ket(NLevelBasis(8), ... )` would fail with `DimensionMismatch`.


AmitRotem:
`iszero` works on operators, but not on states.
So I've added `zero` method for states, similar to the method for operators, so `iszero` would work on states.

amilsted:
Looks good to me!

Issue #97, opened 2023-05-16
Ambiguity of `index` in `ptranspose(rho, index)`


zejian-li:
From the docstring "Partial transpose of rho with respect to subspace specified by index" one might expect the function to perform the partial transpose on the subsystem specified by the index, i.e. if $\hat{\rho} = \hat{A}\otimes\hat{B}$,    
ptranspose($\hat{\rho}$,1) = $\hat{A}^T\otimes \hat{B}$    
instead of    
ptranspose($\hat{\rho}$,1) = $\hat{A}\otimes \hat{B}^T$, which is what the function actually does.

david-pl:
@zejian-li that's a bug. It should actually be the other way around. I never noticed as I only used the `ptranspose` together with the `negativity` where the order of the subsystems doesn't matter. That needs to be fixed and we need some tests for `ptranspose` only. 

Feel free to submit a PR with your changes. Although we'll need to fix the order of the single index version first.

PR #98, opened 2023-05-18
Fix bug in ptranspose and added corresponding tests

zejian-li:
fixed bug reported in #97 
the `ptranspose(rho::DenseOpType{B,B}, indices::Union{Vector{Int},Tuple{Vararg{Int}}})` method is added to support multi-index subsystems, and the original single-index method `ptranspose(rho::DenseOpType{B,B}, index::Int=1)` is kept for compatibility, but implemented as a special case using the multi-index version.
tests dedicated to the two methods are added.

### DIFF
--- a/src/metrics.jl
+++ b/src/metrics.jl
@@ -195,7 +195,7 @@ fidelity(rho::DenseOpType{B,B}, sigma::DenseOpType{B,B}) where {B} = tr(sqrt(sqr
 Partial transpose of rho with respect to subspace specified by indices,    
 where `indices` can be specified by a single integer, an array or a tuple of integers.
 """
-function ptranspose(rho::DenseOpType{B,B}, indices::Union{Vector{Int},Tuple{Vararg{Int}}}) where B<:CompositeBasis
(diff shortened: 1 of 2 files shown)


david-pl:
@zejian-li Thanks for this! It looks fine, I just have one small question: any reason you added another dispatch for `index::Int` ? In Julia, integers are iterable, so the method for `ptranspose(rho::DenseOpType{B,B}, indices::Union{Vector{Int},Tuple{Vararg{Int}}})` also works for `indices::Int` if you just remove the typing, i.e.make the signature `ptranspose(rho::DenseOpType{B,B}, indices)`. Then you can remove the additional `ptranspose(rho::DenseOpType{B,B}, index::Int)` dispatch altogether.

zejian-li:
@david-pl yes you are right! I will change the signature back to `ptranspose(rho::DenseOpType{B,B}, indices=1)`, i.e. the original signature for the current implementation in the master branch, for compatibility.

zejian-li:
Should be good now. The negativity functions will also support now multiple indices without modification.

PR #103, opened 2023-06-09
defined some apply! methods

Abhishek-1Bhatt:
Related to QuantumSavory/QuantumClifford.jl/pull/120. More specifically this pr moves [these](https://github.com/QuantumSavory/QuantumClifford.jl/pull/120/files#diff-77da3915beab8d037a3206102ce2369ae846e9471537eb8c9c6dc45a1f925b8bR71-R91) changes from the previously linked pr to this repo.
cc @Krastanov


Krastanov:
@david-pl this is the first time something needed by QuantumClifford / QuantumSymbolics / QuantumSavory is being submitted to QuantumOptics.

It is the `apply!` function (that we moved from QuantumClifford to QuantumInterface when we made that package). I will go ahead and do a review for @Abhishek-1Bhatt  (he is a SciML gsoc veteran and he is joining my group at UMass Amherst).

However, may we have your blessing for the `apply!` API in principle, not just in the context of this PR?

Krastanov:
@amilsted this overlaps with your lazy embed. When the lazy embed is merged, we should probably investigate whether we can use the same underlying tools for this `apply!`

Krastanov:
@Abhishek-1Bhatt 

- We will probably want to optimize this in the future by using pre-allocated buffers and other improvements. Could you please make sure there are a few simple tests added for this so that we can make the future optimizations without worrying too much? The checks can be along the lines of `apply!(spin_up⊗spin_down, (1,), sigma_x) == spin_down⊗spin_down`.
- What happens when you do `apply!(ket, some_superoperator)`? Usually a superoperator would turn a ket into a density matrix, so the data override will probably fail.
- Could you check whether it gives an appropriate error message when the right and left bases of an operator are different (same for superoperator)? You can make such an operator by doing something like |ket_in_basis_A⟩⟨ket_in_basis_B|

Do not hesitate to message me if some of the quantum information theory behind this is unclear. We can meet to discuss it in more detail.

david-pl:
@Krastanov from my side this is absolutely fine. In general if you need to upstream small-ish bits of code that's fine. The only thing I'd be hesitant of is adding new dependencies upstream (of course that doesn't apply here since there aren't any). So feel free to merge once you're done with your review.

Krastanov:
yup, your suggestion is reasonable!

You can also consider defining `basis(sop::Superoperator) = sop.basis_r` only if all of the supop bases are the same (raising an error otherwise). Same as what is currently happening for `Operator`:

```
julia> b1, b2 = [GenericBasis(i) for i in [4,5]];

julia> s1, s2 = [basisstate(b,1) for b in [b1, b2]];

julia> projector(s1,s1') |> basis
Basis(dim=4)

julia> projector(s1,s2') |> basis
ERROR: QuantumInterface.IncompatibleBases()
```

Abhishek-1Bhatt:
Requires qojulia/QuantumInterface.jl/pull/11

Abhishek-1Bhatt:
Ah, yes that would be in agreement with the output of the other `basis()` dispatch, perfect

Krastanov:
Thank you, this was very helpful!

Issue #99, opened 2023-05-27
compatibility with SciMLOperators


Krastanov:
See https://docs.sciml.ai/SciMLOperators/stable/

It is a parallel development from the SciML organization that has similarities to our Lazy operators and to the waveguide operators from the WaveguideQED.jl extension to QuantumOpticsBase.jl. It would be valuable to understand and document the tradeoffs between our approaches and aim for some level of compatibility.

Krastanov:
Somewhat related to:

- https://github.com/qojulia/QuantumOptics.jl/issues/299
- https://github.com/qojulia/QuantumOptics.jl/issues/298
- https://github.com/qojulia/QuantumOpticsBase.jl/pull/16

Krastanov:
Also related to #172

PR #104, opened 2023-06-12
Time dependent operators

amilsted:
Convenient time-dependence handling for operators. This PR introduces:
* `AbstractTimeDependentOperator`, an interface for adding time-dependence to existing operator types
* `TimeDependentSum` an implementation of time dependence on top of LazySum

Not in this PR: Interface for `QuantumOptics.timeevolution` to allow directly supplying these operators to e.g. `schroedinger_dynamic()`.

This makes it easy to compose time dependent terms without manually maintaining a LazySum, whilst still maintaining efficient dynamical simulation by avoiding unnecessary allocations. For example, one can do:
```julia
b = tensor(FockBasis(10), FockBasis(10))
a1 = embed(b, 1, destroy(b.bases[1]))
a2 = embed(b, 2, destroy(b.bases[2]))

H_static = a1' * a1 + 2 * a2' * a2

H_drive_1 = TimeDependentSum(cos => (a1' + a1))
H_drive_2 = TimeDependentSum((t->cos(2*t)) => (a2' + a2))

H = H_static + 0.1 * H_drive_1 + 0.01 * H_drive_2

# and later
psi0 = basisstate(b, 1)
ts_out, psis_out = timeevolution.schroedinger_dynamic([0.0, 10.0], psi0, H)
```

Time-dependent operators have their own clocks, so they always have a concrete value. The clock can be set with `set_time!(my_op, t)` or `my_op(t)`. Combining operators with different `current_time` is an error.

Time-dependent operators are structured as "wrapper" types that add time-dependence to an existing operator type. A time-dependent sum is only one form. One could also have a time-dependent dense or sparse operator, for example, in whose matrix elements are functions of time.


Krastanov:
@mabuni1998 , this looks similar to some of the helpers you had to develop for WaveguideQED. If you have the bandwidth (after you are done with the writing crunch), your review would be highly valued.

Krastanov:
SciML have been developing something similar (as discussed in https://github.com/qojulia/QuantumOpticsBase.jl/issues/99 ) -- it would be valuable to not become too incompatible with them.

Krastanov:
@amilsted , on a very quick skim through this, it looks really exciting and addressing a pain point I have had. Thank you for making this happen! I will try to provide more detailed review in the next few days.

amilsted:
> SciML have been developing something similar (as discussed in #99 ) -- it would be valuable to not become too incompatible with them.

Oh, wow, there really is a lot of stuff there that is similar to our Lazy and now TimeDependent operators! Looks like, at least functionality-wise, we could already replace LazySum, LazyProduct, and LazyTensor with this stuff and add time-dependence using `update_func` definitions. Seems it would be pretty easy to try a `DataOperator` with the data field set to be a `SciMLOperator` right now. Would be interesting to compare performance and usability!

src/QuantumOpticsBase.jl
```diff
@@ -28,6 +28,11 @@ export Basis, GenericBasis, CompositeBasis, basis,
         #operators_lazytensor
                 LazyTensor, lazytensor_use_cache, lazytensor_clear_cache,
                 lazytensor_cachesize, lazytensor_disable_cache, lazytensor_enable_cache,
+        #time_dependent_operators
+                AbstractTimeDependentOperator, TimeDependentSum, is_const, set_time!,
+                current_time, coefficients, suboperators, eval_coefficient,
```
Krastanov, line 33: I am a bit on the fence about exporting `coefficients`, `suboperators`, and a lot other of these helpers. Could we consider keeping as many of them as possible as private, in case we want to modify something as we are evaluating SciMLOperators?

src/time_dependent_operator.jl
```diff
@@ -0,0 +1,302 @@
+
+import Base: size, *, +, -, /, ==, isequal, adjoint, convert
+
+abstract type AbstractTimeDependentOperator{BL,BR} <: AbstractOperator{BL,BR} end
+
+current_time(::AbstractOperator) = throw(ArgumentError("Time not defined for operator."))
```
Krastanov, line 6: Recently I have started taking cues from the SciML ecosystem and trying to be very detailed in my error messages, suggesting possible workarounds and giving more explanations. This has a very high value in how novices perceive a library. Could you consider expanding this to something like "Time is not defined for this operator. You are probably attempting to use a time-independent operator in a code pathway that expects explicitly time-dependent operators. Consider wrapping ... ...".

src/operators_lazysum.jl
```diff
@@ -153,10 +169,10 @@ end
 identityoperator(::Type{<:LazySum}, ::Type{S}, b1::Basis, b2::Basis) where S<:Number = LazySum(identityoperator(S, b1, b2))
 
 function embed(basis_l::CompositeBasis, basis_r::CompositeBasis, indices, op::LazySum)
-    LazySum(basis_l, basis_r, op.factors, ((embed(basis_l, basis_r, indices, o) for o in op.operators)...,))
+    LazySum(basis_l, basis_r, op.factors, map(o->embed(basis_l, basis_r, indices, o), op.operators))
```
Krastanov, line 172: Any particular reason you prefer an explicit map to an implicit generator comprehension?

test/test_aqua.jl
```diff
@@ -9,6 +9,7 @@ using FillArrays
                   piracy=(broken=true,)
                   )
     # manual piracy check to exclude identityoperator
-    pirates = [pirate for pirate in Aqua.Piracy.hunt(QuantumOpticsBase) if pirate.name != :identityoperator]
+    # FIXME: also adjoint. Move StateVector def to QuantumInterface.
+    pirates = [pirate for pirate in Aqua.Piracy.hunt(QuantumOpticsBase) if pirate.name != :identityoperator && pirate.name != :adjoint]
```
Krastanov, line 13: Is this outdated? I believe `StateVector` is already in `QuantumInterface`. The only new `adjoint` defined here does not seem to be a pirate.

src/time_dependent_operator.jl
```diff
@@ -0,0 +1,302 @@
...
+static_operator(o::AbstractOperator) = o
+
+set_time!(o::AbstractOperator, ::Number) = o
+set_time!(o::LazyOperator, t::Number) = (set_time!.(o.operators, t); return o)
+
+(o::AbstractTimeDependentOperator)(t::Number) = set_time!(o, t)
+
+function _check_same_time(A::AbstractTimeDependentOperator, B::AbstractTimeDependentOperator)
+    current_time(A) == current_time(B) || throw(ArgumentError("Time-dependent operators with different times cannot be combined."))
```
Krastanov, line 15: same comment about error messages

src/time_dependent_operator.jl
```diff
@@ -0,0 +1,302 @@
...
+for func in (:basis, :length, :size, :tr, :normalize, :normalize!,
+    :identityoperator, :one, :eltype, :ptrace)
+    @eval $func(op::AbstractTimeDependentOperator) = $func(static_operator(op))
+end
+
+for func in (:expect, :variance)
+    @eval $func(op::AbstractTimeDependentOperator{B,B}, x::Ket{B}) where B = $func(static_operator(op), x)
+    @eval $func(op::AbstractTimeDependentOperator{B,B}, x::AbstractOperator{B,B}) where B = $func(static_operator(op), x)
+    @eval $func(index::Integer, op::AbstractTimeDependentOperator{B1,B2}, x::AbstractOperator{B3,B3}) where {B1,B2,B3<:CompositeBasis} = $func(index, static_operator(op), x)
+    @eval $func(indices, op::AbstractTimeDependentOperator{B1,B2}, x::AbstractOperator{B3,B3}) where {B1,B2,B3<:CompositeBasis} = $func(indices, static_operator(op), x)
```
Krastanov, line 27: On discourse and slack I have seen some commentary that this type of `eval` can increase compilation and loading time. We are not particularly excessive users of this pattern, so I am not worried, but I thought it is a good idea to mention it for future reference.

src/time_dependent_operator.jl
```diff
@@ -0,0 +1,302 @@
...
+const VecOrTuple = Union{Tuple,AbstractVector}
+
+"""
+    TimeDependentSum(lazysum, coeffs, init_time)
+    TimeDependentSum(::Type{Tf}, basis_l, basis_r; init_time=0.0)
+    TimeDependentSum([::Type{Tf},] [basis_l,] [basis_r,] coeffs, operators; init_time=0.0)
+    TimeDependentSum([::Type{Tf},] coeff1=>op1, coeff2=>op2, ...; init_time=0.0)
+    TimeDependentSum(::Tuple, op::TimeDependentSum)
+
+Lazy sum of operators with time-dependent coefficients. Wraps a `LazySum` `lazysum`,
```
Krastanov, line 42: May we do `@ref` links for these references so that they are linked in the documentation?

Krastanov (approved):
I am pretty excited for this to be merged in the near term, it would be a wonderful addition! I have left some minor comments in, mostly on what is publicly exported and on error messages.

If it is not too burdensome, it would be great to finish the test coverage for the few currently not-covered lines.

Besides these minor notes, this looks good for merging to me!

Krastanov:
The JET failure is seemingly not related to this. I think it is just the usual fluctuation in number of detected issues as JET is continuously being improved.

src/time_dependent_operator.jl
amilsted: Oh, interesting. I thought this was a fairly typical Julian pattern. There are still quite a few examples in the Julia source, but perhaps the disadvantages don't appear for the system image...

src/operators_lazysum.jl
amilsted: Yes - I want it to leave vectors as vectors and tuples as tuples, rather than explicitly constructing a tuple.

src/time_dependent_operator.jl
Krastanov: My vague recollection is that it is heavy on method insertion (which has to run checks for whether anything is being invalidated, which is currently not very optimized). It slows down loading of things that are not in the sysimage, not first execution. As long as it does not create a ton of methods, it is fine, but this pattern makes it very easy to create way too many methods. We are not creating a ton of methods, so we are safe for now.

test/test_aqua.jl
amilsted: `Base.adjoint(a::StateVector) = dagger(a)` is defined in QOB in states.jl. Guess this ought to be in QuantumInterface.

Krastanov: I am weirded out how this did not pop up when I was moving definitions to QuantumInterface. Thanks for taking care of it.

Krastanov: Do you want me to move that out in a separate PR, or should we just merge https://github.com/qojulia/QuantumInterface.jl/pull/14 right now and have the compat section bumped up in this pull request, together with deleting the `adjoint` from QOB?

amilsted: I'm in favor of releasing QI with the adjoint fix. We can also do a separate PR for QOB to remove the duplicate.

amilsted: Oops, accidentally removed the method directly in master. Web VS Code makes this too easy!

Krastanov: Done, release should be out in 15ish minutes. And yes, I really enjoy web VScode and have the same bad habit :D. I will rerun the tests on master after QI is out

amilsted:
> I am pretty excited for this to be merged in the near term, it would be a wonderful addition! I have left some minor comments in, mostly on what is publicly exported and on error messages.
> 
> If it is not too burdensome, it would be great to finish the test coverage for the few currently not-covered lines.
> 
> Besides these minor notes, this looks good for merging to me!

Pretty happy with where this is now, but should perhaps prepare the QO changes at least before releasing this.

amilsted:
Note to self: Perhaps `static_operator` is a misnomer. Maybe `wrapped_operator` instead...

amilsted:
It occurs to me it would be quite a nice generalization to introduce a general `TimeDependentOperator(static_operator, update_function)` that could be easily customized. This is taking inspiration from SciMLOperators, but could also be used to hook up SciML operators with update functions to QO.

amilsted:
I think we could merge this. I'll hold back on the generalized `TimeDependentOperator` for now. Should we bump the minor version, or just make this a patch?

Krastanov:
This is non-breaking, rather just adding new functionality, right? Given that the leading version number is 0, I think the usual Julia prescription is to bump only the patch number. That special treatment of 0-version software is also why a lot of semver folks encourage everyone to go to v1 as soon as possible.

Krastanov:
There are quite a few new JET errors. Could you give them a quick skim, to make sure there is nothing major caused by this addition?

amilsted:
> There are quite a few new JET errors. Could you give them a quick skim, to make sure there is nothing major caused by this addition?

They don't seem to come from this PR. I have tried to fix one anyway :)

amilsted:
> This is non-breaking, rather just adding new functionality, right? Given that the leading version number is 0, I think the usual Julia prescription is to bump only the patch number. That special treatment of 0-version software is also why a lot of semver folks encourage everyone to go to v1 as soon as possible.

Yeah, only reason I hesitate is in case of possible namespace conflicts, but this can be resolved with compat constraints if a quick fix is needed.

Krastanov:
I am in favor of proceeding with the merge. I am not worried about namespace clashes, as user code that defines a `set_time!` will just hide `QuantumOpticsBase.set_time!` from the Main namespace (unless the user also does `import QuantumOpticsBase: set_time!`)

PR #107, opened 2023-06-17
Base.convert(::Type{Operator}, ::Operator)

### LINKED ISSUE #106: `dagger(operator)^2` does not work
```julia-repl
julia> function gen()
           b_fock = FockBasis(10)
           a = destroy(b_fock)
           return dagger(a)^2
       end
gen (generic function with 1 method)

julia> gen()
ERROR: MethodError: Cannot `convert` an object of type 
  Operator{FockBasis{Int64{}},FockBasis{Int64{}},LinearAlgebra.Adjoint{ComplexF64, SparseArrays.SparseMatrixCSC{ComplexF64, Int64}}} to an object of type 
  Operator{FockBasis{Int64{}},FockBasis{Int64{}},SparseArrays.SparseMatrixCSC{ComplexF64, Int64}}

```

Krastanov:
Fixes #106

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "QuantumOpticsBase"
 uuid = "4f57444f-1401-5e15-980d-4471b28d5678"
-version = "0.4.3"
+version = "0.4.4"
 
 [deps]
 Adapt = "79e6a3ab-5dfb-504d-930d-738a2a938a0e"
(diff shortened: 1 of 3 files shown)


Krastanov:
The breakage in QuantumOptics (in the steady state submodule) is not spurious, it is actually there, and it is actually caused by this change... But I have no idea why... Suggestions would be welcomed.

Krastanov:
The breakage was because this new `convert` method was creating copies even when no conversion is necessary (as described in the commit message).

I am merging this as it is strictly a bugfix not affecting the rest of the system and follows Julia base conventions.

PR #108, opened 2023-06-19
Added multiplication of LazyOperators

mabuni1998:
Added functions for multiplying two LazySum operators, LazySum with LazyTensor / LazyProduct and LazyTensor with LazyProduct. I also added tests for the new functions. 

Although debatable whether necessary, there might be situations where you would prefer not to explicitly have to calculate the product and create the appropriate LazyOperator. For example, if you wanted to create a unitary time evolution operator you might write:
```
U = identityoperator(H) - im*dt*H + (-im*dt)^2/2*H^2
```
where H is a LazyOperator. This is currently not possible, but with the few added lines of code here, you can do the above.

One thing to note is that when multiplying a non-lazy operator with a LazyOperator currently the following functions are used:
https://github.com/qojulia/QuantumOpticsBase.jl/blob/f98ec890036ca8de661d877a05471b6e2cfb2912/src/operators_dense.jl#L91-L110

The LazyOperators are thus not contagious with non-lazy operators in multiplication. This is, however, probably for the best since the above lines of code are also used when computing multiplication with a density matrix (which is represented as an operator) and in this instance, we don't want to create a LazyOperator but actually calculate e.g. the expectation value. If we wanted to have contagious LazyOperators in multiplication we would have to distinguish between operators and density matrices.

### DIFF
--- a/src/operators_lazysum.jl
+++ b/src/operators_lazysum.jl
@@ -104,6 +104,44 @@ end
 -(a::O1, b::O2) where {O1<:LazyOperator,O2<:LazyOperator} = LazySum(a) - LazySum(b)
 
 
+#*(a::LazySum{B1,B2}, b::AbstractOperator{B2,B3}) where {B1,B2,B3} = LazySum(a.basis_l, a.basis_r, a.factors, a.operators .* b)
+#*(a::AbstractOperator{B1,B2}, b::LazySum{B2,B3}) where {B1,B2,B3} = LazySum(a.basis_l, b.basis_r, b.factors, a.operators .* b)
+
+#*(a::Operator{B1,B2}, b::LazySum{B2,B3}) where {B1,B2,B3} = LazySum(a.basis_l, b.basis_r, b.factors, a.operators .* b)
+#*(a::LazySum{B1,B2}, b::Operator{B2,B3}) where {B1,B2,B3} = LazySum(a.basis_l, b.basis_r, a.factors, a.operators .* b)
+
+
+function Base.:*(a::LazySum{B1,B2}, b::O2) where {B1,B2,O2<:LazyOperator}
+    c = Array{AbstractOperator}(undef,length(a.operators))
+    for i in eachindex(a.operators)
+        c[i] = LazyProduct(a.operators[i]) * b
+    end
+    LazySum(a.basis_l, b.basis_r, a.factors, (c...,))
+end
+
+function Base.:*(a::O1, b::LazySum{B1,B2}) where {B1,B2,O1<:LazyOperator}
+    c = Array{AbstractOperator}(undef,length(b.operators))
+    for i in eachindex(b.operators)
+        c[i] = a * LazyProduct(b.operators[i])
+    end
+    LazySum(a.basis_l, b.basis_r, b.factors, (c...,))
+end
+
+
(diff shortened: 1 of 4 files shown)


src/operators_lazysum.jl
```diff
@@ -104,6 +104,44 @@ end
...
+#*(a::Operator{B1,B2}, b::LazySum{B2,B3}) where {B1,B2,B3} = LazySum(a.basis_l, b.basis_r, b.factors, a.operators .* b)
+#*(a::LazySum{B1,B2}, b::Operator{B2,B3}) where {B1,B2,B3} = LazySum(a.basis_l, b.basis_r, a.factors, a.operators .* b)
+
+
+function Base.:*(a::LazySum{B1,B2}, b::O2) where {B1,B2,O2<:LazyOperator}
+    c = Array{AbstractOperator}(undef,length(a.operators))
+    for i in eachindex(a.operators)
+        c[i] = LazyProduct(a.operators[i]) * b
+    end
+    LazySum(a.basis_l, b.basis_r, a.factors, (c...,))
```
Krastanov, line 119: You have an array typed to `AbstractOperator`. If your array is already meant to contain abstract types, there is not much reason to do anything more specific than `Any`, especially given that it would be slightly more readable code.

More importantly though, here you create a temporary array and then you immediately destroy it. It is a bit wasteful (causing unnecessary allocations). It might be a bit simpler to just use a "generator" (a "lazy" object that does not actually compute the content until it is queried):

```
julia> generator = (i for i in 1:10)
Base.Generator{UnitRange{Int64}, typeof(identity)}(identity, 1:10)

julia> tuple = (generator...,)
(1, 2, 3, 4, 5, 6, 7, 8, 9, 10)
``` 

So your code can be simplified to something like `LazySum(bl, br, a.factors, ((LazyProduct(op)*b for op in a.operators)...,))`

src/operators_lazysum.jl
```diff
@@ -104,6 +104,44 @@ end
...
+end
+
+function Base.:*(a::O1, b::LazySum{B1,B2}) where {B1,B2,O1<:LazyOperator}
+    c = Array{AbstractOperator}(undef,length(b.operators))
+    for i in eachindex(b.operators)
+        c[i] = a * LazyProduct(b.operators[i])
+    end
+    LazySum(a.basis_l, b.basis_r, b.factors, (c...,))
```
Krastanov, line 127: same simplificaiton here

Krastanov:
I suggested some minor simplifications, but they are more of a style suggestion, no need to actually follow up on them. They will save one allocation each, but this is not a place in the codebase that saving an allocation is very valuable. Especially given that for the third method you have defined, they are not much of a simplification anyway.

Looks good to merge for me! I would also love to have access to this as work is picking up momentum on QuantumSavory and QuantumSymbolics.

Could you also bump the version number to 0.4.4.

I will wait for a couple of days if anyone else wants to comment on this before merging.

Krastanov:
Actually, the bump in Project.toml should be to 0.4.5

Krastanov:
I take the comment about merging in the nearterm back. I think this overlaps a bit with #104 which we should evaluate first, as it brings up a bunch of other functionality as well.

Krastanov:
@mabuni1998 , now that #104 is merged, I think we can close this one as it is a subset of #104. Please let me know if anything is missing from #104 and we can cherry pick it into a new PR.

Issue #109, opened 2023-06-21
regression in `identityoperator()+sparse_matrix` due to the move to `Eye`


Krastanov:
Since 0.4 we have `Eye+sparse=dense`. We should keep the sparsity in situations like this.

Regression reported over slack by Satya Bade:

> Hi Stefan, Thanks, I think the problem is with the addition between Eye and a sparse operator, it gives a dense operator! For example, the following code gives an out of memory error (depending on the available RAM)
>
>     using QuantumOpticsBase
>     bs = SpinBasis(1//2)
>     op = identityoperator(bs) + sigmaz(bs)
>     tensor([op for _ in 1:16]...)
>
> this is possible with the older versions

Krastanov:
This is fixed in https://github.com/qojulia/QuantumOpticsBase.jl/pull/110, but there are still things that can be improved. Namely, the large block of methods with a TODO left in it can be improved by leaning a bit harder into using FillArray and lazy embeds, for scaled identities and for tensor products with identities. Leaving open but renaming

amilsted:
Noting here that FillArrays 1.3 improves on this: https://github.com/JuliaArrays/FillArrays.jl/issues/265

PR #110, opened 2023-06-22
Fix: regression in identityoperator()+sparse_matrix due to the move to Eye #109

SatyaBade12:


test/test_operators_sparse.jl
```diff
@@ -117,6 +117,21 @@ IEye = identityoperator(b_l)
 @test sparse(IEye) == I
 Icomp = identityoperator(b1a) ⊗ identityoperator(b2a) ⊗ identityoperator(b3a)
 @test IEye == Icomp
+for IEye in (IEye, -IEye, IEye, dagger(IEye))
```
amilsted, line 120: IEye is tested twice here?

amilsted:
It seems I probably missed this because `SquareEye` behaves differently to `Eye` under these operations. The former behaves correctly in sums and krons with sparse matrices, the latter does not.

`identityoperator(some_basis)` ought to be producing a SquareEye, but apparently it isn't. In any case, the general Eye case definitely needed fixing!

test/test_operators_sparse.jl
Krastanov: yup, it was supposed to be an adjoint, will fix

Krastanov:
@amilsted , I added tests for non-square operators. If this is green on CI I will merge it and we can then follow up with #111 (which would now have some broken tests)

PR #111, opened 2023-06-22
identityoperator(b) should produce SquareEye

amilsted:
With https://github.com/qojulia/QuantumInterface.jl/pull/15, this allows us to produce identity operators that contain `FillArrays.SquareEye` again.

Worth noting that we could also just choose to do this by checking basis sizes. However, the way basis types are defined right now, this would require runtime dispatch. That said, maybe that's okay for constructing identity operators.

Edit: This is less important with https://github.com/qojulia/QuantumOpticsBase.jl/pull/110. After that is merged, `FillArrays.Eye` ought to properly preserve sparsity.


Krastanov:
This seems good to merge. I reran one of the tests after the merge of #110 

I will merge, check for lingering errors and release. Then we can retest https://github.com/qojulia/QuantumInterface.jl/pull/15 and release.

No need for new compats, right?

Krastanov:
And just to confirm, this new function will not be dispatched on until https://github.com/qojulia/QuantumInterface.jl/pull/15 is merged, right?

Krastanov:
Actually, it seems cleaner to merge https://github.com/qojulia/QuantumInterface.jl/pull/15 first as a breaking release and then bump the compat here. Feel free to proceed yourself with merging both of these, or let me know and I will be happy to do them today -- I will just wait on a confirmation in case I am missing something.

amilsted:
I think this will overwrite the QI method. Indeed, the test have:
```julia
 │  WARNING: Method definition identityoperator(Type{T}, QuantumInterface.Basis) where {T<:Number} in module QuantumInterface at /home/runner/.julia/packages/QuantumInterface/20cdS/src/identityoperator.jl:22 overwritten in module QuantumOpticsBase at /home/runner/work/QuantumOpticsBase.jl/QuantumOpticsBase.jl/src/operators_sparse.jl:72.
│    ** incremental compilation may be fatally broken for this module **
```
So it's okay to release this first, but probably cleaner to do a breaking QI release, as you say. I'm happy to let you do the honors ;)

PR #112, opened 2023-06-28
Use FastExpm.jl for matrix exponential of SparseSuperOperator

akirakyle:
I've found FastExpm to work quite well for creating the time evolution of certain, known to be sparse time evolution superoperators. For example

```julia
b = FockBasis(200)
L = liouvillian(identityoperator(b), [destroy(b)])
exp(L)
```

takes less then a second on my laptop. There's a relevant discussion here https://github.com/JuliaLang/julia/issues/32899 about upstream julia not wanting to implement sparse matrix exponential. However I think it makes sense to rely on FastExpm here since my guess is that most typcial Lindbladians can be computed efficiently by it.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,10 +1,11 @@
 name = "QuantumOpticsBase"
 uuid = "4f57444f-1401-5e15-980d-4471b28d5678"
-version = "0.4.11"
+version = "0.4.12"
 
 [deps]
 Adapt = "79e6a3ab-5dfb-504d-930d-738a2a938a0e"
 FFTW = "7a1cc6ca-52ef-59f5-83cd-3a7055c09341"
+FastExpm = "7868e603-8603-432e-a1a1-694bd70b01f2"
 FillArrays = "1a297f60-69ca-5386-bcde-b61e274b549b"
 LRUCache = "8ac3fa9e-de4c-5943-b1dc-09c6b5f20637"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
@@ -17,6 +18,7 @@ UnsafeArrays = "c4a57d5a-5b31-53a6-b365-19f8c011fbd6"
 [compat]
 Adapt = "1, 2, 3.3"
 FFTW = "1.2"
+FastExpm = "1.1.0"
 FillArrays = "0.13, 1"
 LRUCache = "1"
 QuantumInterface = "0.3.0"
--- a/src/operators_dense.jl
+++ b/src/operators_dense.jl
@@ -226,6 +226,12 @@ function expect(op::DataOperator{B1,B2}, state::DataOperator{B2,B2}) where {B1,B
     result
 end
 
+"""
+    exp(op::DenseOpType)
+
+Operator exponential used, for example, to calculate displacement operators.
+Uses LinearAlgebra's `Base.exp`.
+"""
 function exp(op::T) where {B,T<:DenseOpType{B,B}}
     return DenseOperator(op.basis_l, op.basis_r, exp(op.data))
 end
--- a/src/operators_dense.jl
+++ b/src/operators_dense.jl
@@ -231,6 +231,9 @@ end
 
 Operator exponential used, for example, to calculate displacement operators.
 Uses LinearAlgebra's `Base.exp`.
+
+If you only need the result of the exponential acting on a vector,
+consider using much faster implicit methods that do not calculate the entire exponential.
 """
 function exp(op::T) where {B,T<:DenseOpType{B,B}}
     return DenseOperator(op.basis_l, op.basis_r, exp(op.data))
--- a/src/operators_sparse.jl
+++ b/src/operators_sparse.jl
@@ -53,6 +53,9 @@ Operator exponential used, for example, to calculate displacement operators.
 Uses [`FastExpm.jl.jl`](https://github.com/fmentink/FastExpm.jl) which will return a sparse
 or dense operator depending on which is more efficient.
 All optional arguments are passed to `fastExpm` and can be used to specify tolerances.
+
+If you only need the result of the exponential acting on a vector,
+consider using much faster implicit methods that do not calculate the entire exponential.
 """
 function exp(op::T; opts...) where {B,T<:SparseOpType{B,B}}
     return SparseOperator(op.basis_l, op.basis_r, fastExpm(op.data; opts...))
--- a/src/operators_sparse.jl
+++ b/src/operators_sparse.jl
@@ -1,5 +1,6 @@
 import Base: ==, *, /, +, -, Broadcast
 import SparseArrays: sparse
+import FastExpm: fastExpm
 
 const SparseOpPureType{BL,BR} = Operator{BL,BR,<:SparseMatrixCSC}
 const SparseOpAdjType{BL,BR} = Operator{BL,BR,<:Adjoint{<:Number,<:SparseMatrixCSC}}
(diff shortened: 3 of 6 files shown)


Krastanov:
I am in favor of the FastExpm dependency, it is small and worthwhile.

However, why do this only for sparse superoperators? This seems like it should be appropriate for dense objects too and for operators (not only superoperators) too? This was actually the context in which I had started that Julia discussion a few years ago.

This also would need correctness tests that use the default (slower) exponentiation.

Krastanov:
Maybe I misread the FastExpm documentation? They are for full (rank?) sparse matrices, not dense or sparse matrices, right? Still, why do this only for superoperators, not also for operators?

Krastanov:
And a separate question: KrylovKit is pretty popular in the julia ecosystem. Why use FastExpm instead of KrylovKit? Which one is better for this purpose?

akirakyle:
> However, why do this only for sparse superoperators? This seems like it should be appropriate for dense objects too and for operators (not only superoperators) too?

The short is because that's what I needed for some code I'm writing for my research. I hadn't even realized that exponentials of sparse operators also weren't implemented, but I can add those to this PR as well. I agree that it might make sense to also use the same method for sparse matrics and dense ones as most of the matrix exponentiation libraries seem to take care of checking whether a dense or spares representation will be faster.

> This also would need correctness tests that use the default (slower) exponentiation.

Yes I'll add tests to this PR before merging.

> Maybe I misread the FastExpm documentation? They are for full (rank?) sparse matrices, not dense or sparse matrices, right? Still, why do this only for superoperators, not also for operators?
> And a separate question: KrylovKit is pretty popular in the julia ecosystem. Why use FastExpm instead of KrylovKit? Which one is better for this purpose? https://jutho.github.io/KrylovKit.jl/latest/man/matfun/#KrylovKit.exponentiate

I'm by no means an expert in numerical matrix methods so I may be misunderstanding, however in doing some more research on the methods available, I think that Kroylov methods are not applicable here. It seems Kroylov methods focus on computing the vector result of `exp(A)*x` for a matrix `A` and vector `x`. In my case, I do actually need the full matrix `exp(A)`.

There seems to be several methods for calculating the full matrix exponential. Obviously computing the eigendecomposition is the most straightforward yet most computationally expensive in general. It seems FastExpm essentially just computes the truncated Taylor series approximation to the exponential and thus shouldn't care about the rank of the matrix (although I'd hope it would terminate faster for rank deficient matrices). There is also the "scaling and squaring" based approach which uses the Pade approximant. I'm not sure how this method handles rank deficient matrices.

According to the [`Base.exp` docs](https://docs.julialang.org/en/v1/stdlib/LinearAlgebra/#Base.exp-Tuple{StridedMatrix{var%22#s1397%22}%20where%20var%22#s1397%22%3C:Union{Float32,%20Float64,%20ComplexF32,%20ComplexF64}}), "for symmetric or Hermitian A, an eigendecomposition is used, otherwise the scaling and squaring algorithm is chosen. It looks like [ExponentialUtilities.jl](https://docs.sciml.ai/ExponentialUtilities/stable/matrix_exponentials/) might be the best package to use for the "scaling and squaring" based approach implemented for sparse matrices (it's already a transitive dependency of `QuantumOptics.jl`). I might be able to run some benchmarks between `FastExpm.jl` and `ExponentialUtilities.jl` if that would be useful (or possiblly allow one to chose between all the methods?).

Krastanov:
Yeah, we would definitely need some benchmarking in order to figure out how to proceed.

Of note is that we can pre-allocate cache in the `QuantumOpticsBase.__init__` function that can be used by [ExponentialUtilities.exponential!](https://docs.sciml.ai/ExponentialUtilities/stable/matrix_exponentials/#ExponentialUtilities.exponential!), which probably would be a big win.

It is also perfectly valid to use FastExpm for one data type and ExponentialUtilities for the other.

Let us know if you have the bandwidth to investigate these -- help with this would be greatly appreciated.

akirakyle:
> Let us know if you have the bandwidth to investigate these -- help with this would be greatly appreciated.

It may take me a bit of time to do some benchmarking/investigation here but it would be worthwhile for me to spend some time on it as it is relevant to code I'm writing for my research, where already I'm seeing some preliminary wins over qutip of an order of magnitude or more.

akirakyle:
I started trying to do some benchmarking on this issue, but I seem to have quickly come to the conclusion that `fastExpm` may actually be the only current option.

For calculating the full matrix exponential (as opposed to `exp(tA)*x` for which there appear to be far more packages and algorithms available, but this is not what I'm interested in here) the following are all the implementations I found for julia:

- [`Base.exp`](https://github.com/JuliaLang/julia/blob/d080fe6c9bbc2076a7228d43f2651831ef7f8bac/stdlib/LinearAlgebra/src/dense.jl#L649C8-L649C8)
  - Only works on dense matrices
- [`fastExpm.fastExpm`](https://github.com/fmentink/FastExpm.jl/blob/main/src/FastExpm.jl)
  - Automatically decides whether to use sparse or full as it goes along.
- [`Expokit.padm`](https://github.com/acroy/Expokit.jl/blob/master/src/padm.jl) from 
  - Works on sparse matrices but always returns a dense matrix
- [`ExponentialUtilities.exponential!`](https://github.com/SciML/ExponentialUtilities.jl/blob/30687b1a4d95f2abb7464118efd161614b3ea074/src/exp.jl)
  - Provides several different methods, however none of them worked on sparse arrays. The `ExpMethodGeneric` seems like it should work on sparse arrays but I get a `MethodError: no method matching ldiv!(::LinearAlgebra.Adjoint{Float64, SparseArrays.UMFPACK.UmfpackLU{Float64, Int64}}, ::SparseArrays.SparseMatrixCSC{Float64, Int64})`. [This discussion](https://discourse.julialang.org/t/matrix-division-for-sparse-and-singular-matrices/68253) may be relevant to this error.

Thus for the two options that accept sparse, `fastExpm` will of course be much faster for really sparse operators. As an example:

```julia
using QuantumOpticsBase
import FastExpm: fastExpm
import Expokit: padm

b = FockBasis(49)
L = liouvillian(sparse(identityoperator(b)), [destroy(b)])

@time padm(L.data)
@time fastExpm(L.data)
```

On my machine `padm` takes 8s while `fastExpm` takes 0.04s`. Given this I think it makes sense to continue using `Base.exp` for dense operators and superoperators and add `fastExpm` for sparse operators and superoperators.

Krastanov:
Wonderful work, thank you! I am all on board with using FastExp.

@david-pl , given that this is a new dependency, not even a transitive one, I just want to make you are pinged. It is a small and high-quality dependency, so I will plan to merge this soon.

@akirakyle , any reason the test you added is commented out?

akirakyle:
> @akirakyle , any reason the test you added is commented out?

Because the test above it which it is based on is also commented out, which I'm guessing is due to it relying on `timeevolution` from QuantumOptics and was commented out when QuantumOpticsBase was split out of it. I suppose QuantumOptics could be added as test dependency, but I'm not if this is possible with it creating a circular dependency?

Krastanov:
@akirakyle , it seems some of the new lines that are calls to fastExp are not currently under tests (probably because you were busy making valuable benchmarks :D ) -- now that you have proven this choice is the best, would you be able to do some final polish and add tests to cover the new functionality?

Krastanov:
Your explanation for the commented out test makes perfect sense. Do not worry about changing that one (sorry for the out of order responses, I did not see you had already answered that)

Krastanov:
Something like `exp(sparse(displacement_hamiltonian)) * displaced_ket == different_displaced_ket` or `exp(sparse(operator)).data == exp(operator.data)` can be a good test of this functionality without relying on `timeevolution`

akirakyle:
I've added some tests, updated the docstrings, and enabled passing the optional arguments to `fastExpm` which control its precision. Hopefully this is now ready to merge?

Krastanov:
This is great, thank you! It will be merged and released shortly!

I added a small piece of advice to the docstrings. Let me know if you disagree with it and I can modify it.

akirakyle:
> I added a small piece of advice to the docstrings. Let me know if you disagree with it and I can modify it.

Thanks for getting this merged! The advice you added looks good to me! I wonder if it might be worthwhile suggesting packages? In some of my testing `expv` from `ExponentialUtilities` was faster than `exponentiate` from `KrylovKit` but each package has many options that can probably be tuned for an application.

PR #113, opened 2023-06-28
Add sprepost to superoperators

akirakyle:
`sprepost` is necessary for creating superoperators with different input and output hilbert space dimensions since `spre(A)*spost(B) != sprepost(A, B)` in such cases.

Note that I'm not sure exactly what the sensible behavior should be when calling `spre` or `spost` with an operator that doesn't have the same `basis_l` and `basis_r` and I've opened a relevant issue with qutip (https://github.com/qutip/qutip/issues/2171).

### DIFF
--- a/test/test_apply.jl
+++ b/test/test_apply.jl
@@ -43,12 +43,13 @@ b2 = FockBasis(2)
 k1 = coherentstate(b1, 0.39)
 k2 = coherentstate(b2, 1.4)
 op = projector(k1, k2')
-sOp2 = spre(op)
 
 try
+    sOp2 = spre(op)
     QuantumOpticsBase.apply!(st, 1, sOp2)
 catch e
-    @test typeof(e) <: QuantumInterface.IncompatibleBases
+    @test typeof(e) <: ArgumentError
+    #@test typeof(e) <: QuantumInterface.IncompatibleBases
 end
 
 end #testset
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "QuantumOpticsBase"
 uuid = "4f57444f-1401-5e15-980d-4471b28d5678"
-version = "0.4.10"
+version = "0.4.11"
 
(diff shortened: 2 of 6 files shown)


Krastanov:
Please excuse the long delay on this, I wanted to find some time to read through your qutip issue.

Any chance you can add to the tests something that verifies the `spre(A)*spost(B) != sprepost(A, B)` case you mentioned, just so that if someone in the future decides to do some optimizations, they do not inadvertently break this edge case? Otherwise it looks good to merge.

test/test_apply.jl
```diff
@@ -36,19 +36,4 @@ st = coherentstate(FockBasis(1), 1.4)
...
-k1 = coherentstate(b1, 0.39)
-k2 = coherentstate(b2, 1.4)
-op = projector(k1, k2')
-sOp2 = spre(op)
-
-try
-    QuantumOpticsBase.apply!(st, 1, sOp2)
-catch e
-    @test typeof(e) <: QuantumInterface.IncompatibleBases
-end
```
Krastanov, line 52: Any reason to remove this test? Should it not simply test for the new error you have?

akirakyle: The test wouldn't be for `apply!` anymore since the error is now thrown when calling `spre(op)` which is already tested in `test_superoperators.jl` so I didn't think it made sense have such a test in this file.

Krastanov: Let's still keep them -- down the line something might be modified further, and a mistake might be caught only by these tests. Generally, avoid removing tests, even if they are redundant -- at some point they might not have been redundant and we do not want to risk further refactoring in the future making them necessary again.

akirakyle: I just reworked this test to now pass but without removing it. I left the previously tested for error commented out in the spirit of avoiding removing tests.

Krastanov:
Thank you for implementing and polishing this!

PR #114, opened 2023-06-28
Add squeezing operator

akirakyle:


Krastanov:
@akirakyle, a couple of suggestions on this:

- [ ] given the work you are doing in https://github.com/qojulia/QuantumOpticsBase.jl/pull/112, should this use a better non-dense exponentiation? Maybe we should wait on #112 before deciding how to proceed here?
- [ ] could you add in the docstring the latex math expression for the formula being computed, so that people do not need to worry whether there is a minus sign or a factor of 1/2 somewhere?

akirakyle:
>     given the work you are doing in [Use FastExpm.jl for matrix exponential of SparseSuperOperator #112](https://github.com/qojulia/QuantumOpticsBase.jl/pull/112), should this use a better non-dense exponentiation? Maybe we should wait on [Use FastExpm.jl for matrix exponential of SparseSuperOperator #112](https://github.com/qojulia/QuantumOpticsBase.jl/pull/112) before deciding how to proceed here?

Like `displace` which is always nonzero everywhere, `squeeze` always nonzero in half the entries, so I think a dense representation will always be more efficient. However, it may be that using one of the other methods discussed in #112 would be faster than `base.exp` which is using an eigendecomposition according to its docstring. If I'm going to do some benchmarking, it would be worthwhile to include these two cases as they may warrant a different method than for exponentiation a typical lindbladian.

Krastanov:
Yup, that is a good reason to merge as is. We can investigate improvements down the line after we know more from #112. This will be merged shortly.

PR #116, opened 2023-06-29
Add identitysuperoperator and dagger for SuperOperator

akirakyle:
I'm not sure that I've implemented these in the right way as I'm still getting the hang of Julia's type system. So I'm opening this as a draft PR with the hope that maybe I can get some feedback on how I should actually implement these, as I am needing them for my research. My guess is that maybe superoperators need to have types similar to `DenseOpAdjType` and `SparseOpAdjType` implemented  so that adjoints can be done lazily?

### DIFF
--- a/src/QuantumOpticsBase.jl
+++ b/src/QuantumOpticsBase.jl
@@ -3,7 +3,7 @@ module QuantumOpticsBase
 using SparseArrays, LinearAlgebra, LRUCache, Strided, UnsafeArrays, FillArrays
 import LinearAlgebra: mul!, rmul!
 
-import QuantumInterface: dagger, directsum, ⊕, dm, embed, expect, identityoperator,
+import QuantumInterface: dagger, directsum, ⊕, dm, embed, expect, identityoperator, identitysuperoperator,
         permutesystems, projector, ptrace, reduced, tensor, ⊗, variance, apply!, basis, AbstractSuperOperator
 
 # index helpers
--- a/src/QuantumOpticsBase.jl
+++ b/src/QuantumOpticsBase.jl
@@ -34,6 +34,7 @@ export Basis, GenericBasis, CompositeBasis, basis,
         #superoperators
                 SuperOperator, DenseSuperOperator, DenseSuperOpType,
                 SparseSuperOperator, SparseSuperOpType, spre, spost, liouvillian,
+                identitysuperoperator,
         #fock
                 FockBasis, number, destroy, create,
                 fockstate, coherentstate, coherentstate!,
--- a/src/superoperators.jl
+++ b/src/superoperators.jl
@@ -97,6 +97,9 @@ end
 -(a::SuperOperator) = SuperOperator(a.basis_l, a.basis_r, -a.data)
 -(a::SuperOperator, b::SuperOperator) = throw(IncompatibleBases())
 
+identitysuperoperator(b::Basis) =
+    SuperOperator((b,b), (b,b), Eye{ComplexF64}(length(b)^2))
+
 identitysuperoperator(op::DenseSuperOpType) = 
     SuperOperator(op.basis_l, op.basis_r, Matrix(one(eltype(op.data))I, size(op.data)))
 
--- a/src/superoperators.jl
+++ b/src/superoperators.jl
@@ -97,6 +97,16 @@ end
 -(a::SuperOperator) = SuperOperator(a.basis_l, a.basis_r, -a.data)
 -(a::SuperOperator, b::SuperOperator) = throw(IncompatibleBases())
 
+identitysuperoperator(op::DenseSuperOpType) = 
+    SuperOperator(op.basis_l, op.basis_r, Matrix(one(eltype(op.data))I, size(op.data)))
+
+identitysuperoperator(op::SparseSuperOpType) = 
+    SuperOperator(op.basis_l, op.basis_r, sparse(one(eltype(op.data))I, size(op.data)))
+
+dagger(x::DenseSuperOpType) = SuperOperator(x.basis_r, x.basis_l, copy(adjoint(x.data)))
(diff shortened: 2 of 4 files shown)


Krastanov:
@akirakyle , thank you for the numerous submissions you have made over the last two days! It will not be immediate, but over the next 7 days we should be able to provide feedback on your contributions and help getting them merged! Your work on all of these is very much appreciated!

Krastanov:
@amilsted , as I was reviewing this, I noticed the following in the behavior of `identityoperator` which surprised me:

```
julia> b = FockBasis(10)
Fock(cutoff=10)

julia> identityoperator(b) |> typeof
Operator{FockBasis{Int64}, FockBasis{Int64}, LinearAlgebra.Diagonal{ComplexF64, FillArrays.Ones{ComplexF64, 1, Tuple{Base.OneTo{Int64}}}}}

julia> identityoperator(create(b)) |> typeof
Operator{FockBasis{Int64}, FockBasis{Int64}, SparseArrays.SparseMatrixCSC{ComplexF64, Int64}}

julia> identityoperator(dense(create(b))) |> typeof
Operator{FockBasis{Int64}, FockBasis{Int64}, Matrix{ComplexF64}}
```

Could you confirm that it is on purpose that `identityoperator(op::OperatorType)` returns an operator of type `OperatorType`? I can see arguments for that and I can see arguments for **not** preserving the operator type and rather always returning `Eye` data, so I wanted to make sure.

Krastanov:
@akirakyle , this looks great! Could you make the following two additions?

- [ ] a `identitysuperoperator(b::Basis)` method that uses `Eye` similarly to [the square `identityoperator`](https://github.com/qojulia/QuantumOpticsBase.jl/blob/ab9995fa7e62487b4c6ae42dbc61f7f4fa1fee78/src/operators_sparse.jl#L68) - the `Eye` object has a lot of extra niceties available to it than a typical sparse operator
- [ ] add a docstring to `identitysuperoperator`
- [ ] a few tests of the functionality you have created along the lines of:

```
b = GenericBasis(2)⊗GenericBasis(3)
i1 = identitysuperoperator(b)
i2 = identitysuperoperator(sparse(spre(create(b))))
i3 = identitysuperoperator(dense(spre(create(b))))
@test i1*state == i2*state == i3*state == state == dagger(i1)(state) == ...
```

Lastly, we should probably declare a method-less `identitysuperoperator` in [`QuantumInterface`](https://github.com/qojulia/QuantumInterface.jl/blob/main/src/identityoperator.jl) and import and extend it here. We use `QuantumInterface` to declare functions that other libraries might want to extend without depending on `QuantumOptics`, e.g. [`QuantumClifford`](https://github.com/Krastanov/QuantumClifford.jl) and [`QuantumSymbolics`](https://github.com/QuantumSavory/QuantumSymbolics.jl) -- if that seems too convoluted to deal with right now, please just go ahead with the rest of the changes and I will take care of `QuantumInterface`.

Thank you for the valuable work you are contributing!

amilsted:
> Could you confirm that it is on purpose that `identityoperator(op::OperatorType)` returns an operator of type `OperatorType`? I can see arguments for that and I can see arguments for **not** preserving the operator type and rather always returning `Eye` data, so I wanted to make sure.

Yeah, this is how it has worked for a long time. I suppose it works analogously to `one()` on `AbstractArray` subtypes. Indeed, `one()` on operator types calls `identityoperator`. I agree this is not obviously the best outcome in all circumstances.

akirakyle:
> `identitysuperoperator(b::Basis)` method that uses `Eye` similarly to [the square `identityoperator`](https://github.com/qojulia/QuantumOpticsBase.jl/blob/ab9995fa7e62487b4c6ae42dbc61f7f4fa1fee78/src/operators_sparse.jl#L68) - the `Eye` object has a lot of extra niceties available to it than a typical sparse operator
> 
> Lastly, we should probably declare a method-less `identitysuperoperator` in [`QuantumInterface`](https://github.com/qojulia/QuantumInterface.jl/blob/main/src/identityoperator.jl) and import and extend it here. 

Just so I understand correctly before working on this, should I create a `identitysuperoperator.jl` file in `QuantumInterface` which mirrors the set of functions in `identityoperator.jl` then implement `identitysuperoperator` here in a similar way which does method dispatch based on a `::Type{T}` argument called from `QuantumInterface`?

Krastanov:
> Just so I understand correctly before working on this, should I create a `identitysuperoperator.jl` file in `QuantumInterface` which mirrors the set of functions in `identityoperator.jl` then implement `identitysuperoperator` here in a similar way which does method dispatch based on a `::Type{T}` argument called from `QuantumInterface`?

I would suggest something much simpler as a start. Just declare a `function identitysuperoperator end` without any method definitions in `QuantumInterface` and import it here. That way `QuantumClifford` also can use `identitysuperoperator` for its own types without depending on `QuantumOpticsBase`. Down the line we might want something more sophisticated, but I prefer we do the simpler fast thing first.

Krastanov:
An `identitysuperoperator` was added to QuantumInterface 0.3.1 so now we should be able to just import it from there. 

QuantumInterface 0.3.1 registration: https://github.com/JuliaRegistries/General/pull/87597

Krastanov:
@akirakyle , I made a small change to use the `identitysuperoperator` from the `QuantumInterface` namespace, so that this can also be cohesively used with `QuantumClifford` and company. If you have the bandwidth, could you tackle the two outstanding requests:

- `identitysuperoperator(::Basis)` that internally uses the sparse `Eye` type
- Some test of functionality, to cover all the new additions

akirakyle:
Thanks for adding `identitysuperoperator` to `QuantumInterface`. I've now added a `identitysuperoperator(::Basis)` method and the simplest of tests.

akirakyle:
Just realized I forgot to add tests for `dagger`, will do here shortly...

akirakyle:
Simple tests added for `dagger`

PR #122, opened 2023-07-21
Fix displace_analytical when alpha is an integer type

akirakyle:
I recently got bitten by this subtle bug caused when naively calling `displace_analytical(FockBasis(30), 5)` which then causes `alpha^(n - m)` to overflow for `n-m>27` since it's calculated as `Int64`. I think given how easy it is to overflow for even smallish displacements, we should ensure displace_analytical always computes using floating point.


Krastanov:
Thank you, @akirakyle !

PR #133, opened 2023-09-09
Make `apply!` work when used with any order of subsystem indices

Abhishek-1Bhatt:
`apply!` needs to be able to permute the operator based on the order of the indices of the qubits/subsystems, e.g., doing CNOT<sub>2->1</sub>. Resolves QuantumSavory/QuantumSavory.jl/issues/39 cc @Krastanov

### DIFF
--- a/src/apply.jl
+++ b/src/apply.jl
@@ -31,6 +31,9 @@ function apply!(state::Ket, indices, operation::T) where {T<:AbstractSuperOperat
 end
 
 function apply!(state::Operator, indices, operation::T) where {T<:AbstractSuperOperator}
+    if length(indices)>1
+        throw("Applying SuperOperator to multiple qubits/operators is not supported currently, due to missing tensor product method for SuperOperators")
+    end
     op = basis(state)==basis(operation) ? operation : embed(basis(state), indices, operation)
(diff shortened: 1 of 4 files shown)


Abhishek-1Bhatt:
Would it be better to just have an optional boolean keyword arg in `apply!` like `perm` instead of running the for loop for every multi-qubit system?

Krastanov:
Thanks for starting this!

The apply! on Operator, and the apply! with SuperOperator both suffer from the same bug and should be fixed.

I would suggest not adding any keyword arguments to `apply!`. For now we can just pay the cost of this inefficient implementation. The proper fix for the slowness would be to use a sufficiently general stridded array multiplication.

Krastanov:
Also, I would suggest having a test that has three operators and a permutation that is not its own inverse, e.g. `apply!(a⊗b⊗c, (2,3,1),A⊗B⊗C) ≈ (C*a)⊗(A*b)⊗(B*c)`. The current CNOT tests do not cover such cases (because permutations of 2 elements are their own inverses).

Krastanov:
I did some minor cleanup in the last commit. Thanks for taking care of this!

PR #136, opened 2023-09-12
Fix sparse matrix exponential of zero since FastExpm doesn't handle it

akirakyle:


Krastanov:
All tests pass, the coverage goes up, fixes a regression, and the code is straightforward. The only worry is that the function is not type stable anymore, but this seems like a situation where that is reasonable (an Eye matrix is a nicer result to deal with than a sparse matrix for downstream consumers).

Thank you for taking care of this!

PR #144, opened 2023-09-18
Fixes and tests for `apply!`

Krastanov:
A previous pull request fixed a few issues with apply!, but there were still a number of edge cases that were not covered. This addresses such cases, simplifies the checks for whether an embedding is necessary, and properly detects permutations of the subsystem indices.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "QuantumOpticsBase"
 uuid = "4f57444f-1401-5e15-980d-4471b28d5678"
-version = "0.4.15"
+version = "0.4.16"
 
 [deps]
 Adapt = "79e6a3ab-5dfb-504d-930d-738a2a938a0e"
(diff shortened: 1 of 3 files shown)


Krastanov:
@Abhishek-1Bhatt , could you review this, as it addresses code you worked on?

I also modified some of your tests to use `@test_throws` instead of a try-catch. The try-catch is not reliable here because one day the methods could have just stopped throwing and the tests would not have caught that.

It does not use the expensive `permutesystems` anymore, but more importantly, it fixes the following regression from the previous pull request: `apply!(state_with_3_subsystems, [1,2], operator_with_2_subsystems)` was broken because it was calling the meaningless `permutesystems(operator_with_3_subsystems, [1,2])`.

Abhishek-1Bhatt:
This looks perfect! Looks like I also missed the fact that `embed` can permute the system for us given the indices. The issue with the previous version was that we were only calling `embed` when the `basis` for state and operator were not equal, but it also needed to be called when its equal but the order of indices isn't ascending. 
 I see that `is_apply_shortcircuit` fixes that.

PR #153, opened 2024-01-25
Add a dispatch on dagger for Eye to keep sparsity

david-pl:
@Krastanov this should fix the test failures in #152.

FYI, @amilsted & @SatyaBade12 I had to add another method here, otherwise tests fail on Julia v1.10

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 /docs/build/
 Manifest.toml
+.vscode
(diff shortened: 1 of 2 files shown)


david-pl:
Not sure about the CI failures on 1.6 now (they also seem to be caused by `Eye`), but I'm gonna go ahead and merge this to fix the builds.

Issue #159, opened 2024-04-12
`embed(::NotCompositeBasis, 1, operator)`


Krastanov:
I have some generic code looking like this:

```
embed(SpinBasis(1//2)^n, i, gate)
```

 that occasionally at runtime ends up being `embed(SpinBasis(1//2), i, gate)` which fails.

`embed` expects a composite basis and `basis^1` autosimplifies to `basis`, so the generic code above fails.

We need a small stub distpatch for these "do not operation" edge cases.

a-eghrari:
My first impression is a simple dispatch like `embed(b::Basis, indices, ops) = ops` is enough. Am I right?

Krastanov:
Yes, but I am worried something as simple would cause surprises down the line. Having `indices == 1 || indices == (1,) || indices == [1]` with a clear error message would be safer.

Also, `ops` should probably be typed, to avoid type piracy against `QuantumInterface.jl`

PR #162, opened 2024-05-13
Handle single basis in embed

a-eghrari:
This commit addresses the issue where the `embed` function fails when it receives a single basis instead of a composite basis as input. #159

### DIFF
--- a/src/operators.jl
+++ b/src/operators.jl
@@ -103,6 +103,20 @@ function embed(basis_l::CompositeBasis, basis_r::CompositeBasis,
 end
 
 """
+    embed(basis::Basis, indices, ops::AbstractOperator)
+
+Special case of the `embed` to handle embedding an operator `ops` into a single
+basis.
+"""
+function embed(basis::Basis, indices, ops::AbstractOperator)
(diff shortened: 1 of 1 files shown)


Krastanov:
I only now noticed I have not enabled the tests for your PR. Apologies for that, Github does not run tests automatically on first PR as a security precaution.

Running the tests reveals that it would be best to make this change in the foundational package `QuantumInterface.jl`, otherwise we have a couple of issues:

- This change will not be picked up by other libraries that might benefit from it
- It causes type piracy given that `Basis` is not owned by this package, but it is owned by `QuantumInterface.jl`

Could you consider making the contribution to that other package?

Also, to avoid ambiguity, it should probably be defined for `embed(basis::Basis, indices, ops::AbstractOperator)` (i.e. there should be a type assert on the last argument).

Thank you again for the contribution. Really sorry for making you wait so long for a review.

Krastanov:
I see you have made a PR to QuantumInterface at https://github.com/qojulia/QuantumInterface.jl/pull/24
Thank you! I will close this PR here and proceed there.

PR #169, opened 2024-06-20
Export and document `static_operator`

amilsted:
`static_operator()` is useful for getting a non-lazy matrix representation of a time-dependent operator, for example in order to diagonalize it at a time `t`.


Krastanov:
LGTM

PR #170, opened 2024-06-20
Add a charge basis

amilsted:
This basis is a truncated U(1) irrep basis intended to represent quantized charge. The charge operator is defined as are the conjugate (shift) operators `e^(ikφ)` associated with the U(1) degree of freedom φ.

Could do a minor version bump after this as we will export a bunch of new functions.

### DIFF
--- a/docs/src/api.md
+++ b/docs/src/api.md
@@ -307,6 +307,31 @@ coherentstate
 coherentstate!
 ```
 
+### [Charge](@id API: Charge)
+
+```@docs
+ChargeBasis
+```
+
+```@docs
+ShiftedChargeBasis
+```
+
+```@docs
+chargeop
+```
+
+```@docs
+expiφ
+```
+
+```@docs
+cosφ
+```
+
+```@docs
+sinφ
+```
 
 ### [N-level](@id API: N-level)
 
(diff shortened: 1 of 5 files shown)


Krastanov:
This looks good to me, but I would really appreciate an example of an interesting Hamiltonian for a system of actual interest dropped somewhere in a docstring, if you have the bandwidth for this. Otherwise I feel someone would see it in the documentation but struggle to think of a situation where they would want to use it.

Krastanov:
I restarted one of the tests to see whether codecov would pick things up now that the token is properly set.

amilsted:
@Krastanov I have added a transmon Hamiltonian as an example (also hooked up the docs).

Krastanov:
Thanks! I am in favor of merging!

amilsted:
@david-pl Are you happy with this?

amilsted:
@Krastanov Any objections to a minor version bump?

Krastanov:
not at all -- you never need to ask me that, I am always in favor of quick releases after each PR

amilsted:
Right, I just meant minor vs. patch. If we do minor, we'll need to bump compat on QO.jl too, but I think it's worth doing when the exported functions change a bunch. For example, this will help me migrate private code to use the charge basis.

Krastanov:
Apologies, I misunderstood. Yes, I believe your plan makes sense.

amilsted:
Ok, great. I will also bump the QO.jl minor version, since it reexports QOB.

PR #171, opened 2024-07-04
Eliminate allocs in vector-based TimeDependentSums

amilsted:
This is required to avoid allocs in set_time! with vector-backed TimeDependentSums.
Supports https://github.com/qojulia/QuantumOptics.jl/pull/401.

### DIFF
--- a/src/time_dependent_operator.jl
+++ b/src/time_dependent_operator.jl
@@ -160,14 +160,8 @@ function set_time!(o::TimeDependentSum, t::Number)
         o.current_time = t
         update_static_coefficients!(static_operator(o), coefficients(o), t)
     end
-    subops = suboperators(o)
-    if subops isa Tuple
-        set_time!.(subops, t)
-    else
-        for so in subops
-            set_time!(so, t)
-        end
-    end
+    # foreach is type-stable for tuples and concrete-typed vectors
+    foreach(o -> set_time!(o, t), suboperators(o))
     o
 end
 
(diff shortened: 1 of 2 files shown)


src/time_dependent_operator.jl
```diff
@@ -160,7 +160,14 @@ function set_time!(o::TimeDependentSum, t::Number)
         o.current_time = t
         update_static_coefficients!(static_operator(o), coefficients(o), t)
     end
-    set_time!.(suboperators(o), t)
+    subops = suboperators(o)
+    if subops isa Tuple
+        set_time!.(subops, t)
+    else
+        for so in subops
```
david-pl, line 167: @amilsted maybe a stupid questiom, but isn't that for loop also stable for tuples?

And another one: if we do need the two `if` branches, wouldn't it be better to create a helper function with two methods and let multiple dispatch handle things? That would at least avoid the dynamic check. Not sure if it makes a difference, though. Chances are the compiler figures this out anyway.

amilsted: Yeah, the type of `subops` ought to be inferrable, so this branch should not happen at run time. Maybe worth checking.

As for the loop, the loop variable will not have a concrete type in case of an inhomogenous tuple, so I think this would lead to runtime dispatch. Again, I can check.

david-pl: Also not sure about the branching, but it's probably fine.

> As for the loop, the loop variable will not have a concrete type in case of an inhomogenous tuple

Ah, you're right, my bad. Thanks for the answer!

PR #180, opened 2024-11-24
Remove RandomMatrices as a dependency

apkille:
It looks like RandomMatrices is only used in this package to generate Haar distributed unitary matrices within the methods `randstate_haar` and `randunitary_haar`. This PR implements a slightly faster algorithm that roughly halves the number of allocations and gets rid of RandomMatrices as a dependency.

Before:
```
julia> for i in 10:10:50
          b = SpinBasis(i)
          @btime randunitary_haar($b)
       end
  25.125 μs (41 allocations: 64.75 KiB)
  88.458 μs (41 allocations: 233.47 KiB)
  192.958 μs (41 allocations: 479.50 KiB)
  415.917 μs (41 allocations: 813.44 KiB)
  638.834 μs (41 allocations: 1.21 MiB)
```
After:
```
julia> for i in 10:10:50
           b = SpinBasis(i)
           @btime randunitary_haar($b)
       end
  23.958 μs (23 allocations: 49.94 KiB)
  84.666 μs (23 allocations: 179.42 KiB)
  185.708 μs (23 allocations: 361.56 KiB)
  397.792 μs (23 allocations: 606.28 KiB)
  598.959 μs (23 allocations: 913.59 KiB)
```


Krastanov:
huh... who would have thought that RandomMatrices is not already excessively optimized.

Do you mind also checking if `qr!` makes things even faster?

Either way, this is great -- it also removes the transitive dependency on a bunch more packages. Should be merged pretty soon

src/state_definitions.jl
```diff
@@ -16,7 +15,14 @@ randstate(b) = randstate(ComplexF64, b)
...
 Returns a Haar random pure state of dimension `length(b)` by applying a Haar random unitary to a fixed pure state.
 """
-randstate_haar(b::Basis) = Ket(b, rand(Haar(2), length(b), 2)[:,1])
+function randstate_haar(b::Basis)
+    dim = length(b)
+    mat = rand(ComplexF64, dim, dim)
+    q, r = qr(mat)
+    d = Diagonal([r[i, i] / abs(r[i, i]) for i in 1:dim])
+    data = q * d
+    return Ket(b, data[:,1])
```
Krastanov, line 24: Isn't this just the first column of q up to some random global phase? We can probably skip line 22 and 23 if that is the case, making this even faster.

apkille:
> huh... who would have thought that RandomMatrices is not already excessively optimized.

Yeah, I was surprised by this too. I was planning on using RandomMatrices for Gabs but was shocked by the number of allocations it was producing. It also seems like RandomMatrices has been dealing with a maintenance issue for quite a while...

> Do you mind also checking if `qr!` makes things even faster?

Checked same benchmark as before and `qr!` gives nearly the same exact speed with a few less allocations, so I made the change.

> Either way, this is great -- it also removes the transitive dependency on a bunch more packages. Should be merged pretty soon

PR #185, opened 2025-03-07
Importing metrics from QuantumInterface.jl

Hamiltonian-Action:
Abstract definition to be shifted towards QuantumInterface.jl for suitable implementation amongst the several dependant packages.


Krastanov:
ah... to avoid piracy we need to also move this method:
```
entropy_vn(psi::StateVector; kwargs...)
```

as both `entropy_vn` and `StateVector` are defined now in `QunatumInterface`

Krastanov:
in particular, it is this definition `entropy_vn(psi::StateVector; kwargs...) = entropy_vn(dm(psi); kwargs...)`, assuming `dm` itself is defined in `QuantumInterface`

The other option is to added to the whitelist of pirates in the Aqua tests in this repository

I do not have a strong preference... Lightly leaning towards it being moved to `QuantumInterface`

Krastanov:
Ah, you are making a very good point.

There is an under-specified part of the API we should worry about: in monte carlo quantum trajectories we work with unnormalized states. I am not sure what would be the least-confusing way to set those up with entropy_vn.

If not for those, I would agree that you should just set entropy_vn=0 for state vectors.

Now the question is, can we imagine any reason anyone would have ever called entropy_vn on a non-normalized state. If we can not come up with a reason, I am happy to proceed with your suggestion, even though strictly speaking it is a breaking change.

Krastanov:
Thinking through this, I have trouble imagining a correct use of `S(not_normalized_prob_distribution)`. The entropy is not a linear function, so my example of its use in Monte Carlo quantum trajectories feels to me as a stretch (because you can not then do a weighted sum over each trajectory).

I think my preference is to go for your initial suggestion of just returning `0`. If you have the time/inclination and feel the change is not unreasonable, feel free to proceed with that change in QuantumInterface (and just bump the version/changelog so we can immediately release it).

Project.toml
```diff
@@ -25,7 +25,7 @@ FastGaussQuadrature = "0.5, 1"
 FillArrays = "0.13, 1"
 LRUCache = "1"
 LinearAlgebra = "1"
-QuantumInterface = "0.3.3"
+QuantumInterface = "0.3.9"
```
akirakyle, line 28: Any reason for the bump from 3 to 9?

akirakyle (approved):

Project.toml
akirakyle: Right! Sorry for the noise, I left this comment before looking at the PR over at QuantumInterface

apkille:
@krastanov ping in case you missed this.

PR #188, opened 2025-04-07
Remove usage of `PauliBasis`

akirakyle:
Should be merged in conjunction with https://github.com/qojulia/QuantumInterface.jl/pull/52


test/test_pauli.jl
```diff
@@ -5,11 +5,10 @@ using QuantumOpticsBase
...
 @testset "pauli" begin
 
-@test_throws MethodError PauliBasis(1.4)
-
+b = SpinBasis(1//2)
 # Test conversion of unitary matrices to superoperators.
-q2 = PauliBasis(2)
-q3 = PauliBasis(3)
+q2 = b^2
+q3 = b^3
```
Krastanov, line 11: `PauliBasis` is just being renamed, not completely removed, right? There is still value to having a convenient way to represent multi-qubit basis. This probably should just use the new name. Same for the other changes.

Krastanov (changes requested):
Similar to the comments in the QuantumInterface PR, this is useful functionality that should not just be removed. I agree the name was poorly chosen, but we can deprecate the current name and just use a new name for the same capability (e.g. `QubitBasis` or something like that).

PR #193, opened 2025-04-10
Add generalized pauli operators (aka clock and shift matrices)

### LINKED ISSUE #192: Add generalized Pauli operators
Add the generalized Pauli operators characterized by the Weyl commutation relations, also called the clock and shift matrices. See, for example:
- https://en.wikipedia.org/wiki/Generalizations_of_Pauli_matrices#Construction:_The_clock_and_shift_matrices
- https://www.pnas.org/doi/abs/10.1073/pnas.46.4.570
- https://link.springer.com/chapter/10.1007/3-540-49208-9_27

akirakyle:
Closes #192

### DIFF
--- a/src/QuantumOpticsBase.jl
+++ b/src/QuantumOpticsBase.jl
@@ -58,7 +58,7 @@ export Basis, GenericBasis, CompositeBasis, basis,
                 PositionBasis, MomentumBasis, samplepoints, spacing, gaussianstate,
                 position, momentum, potentialoperator, transform,
         #nlevel
-                NLevelBasis, transition, nlevelstate,
+                NLevelBasis, transition, nlevelstate, paulix, pauliz, pauliy,
         #manybody
                 ManyBodyBasis, FermionBitstring, fermionstates, bosonstates,
                 manybodyoperator, onebodyexpect,
--- a/src/nlevel.jl
+++ b/src/nlevel.jl
@@ -31,3 +31,65 @@ function nlevelstate(::Type{T}, b::NLevelBasis, n::Integer) where T
     basisstate(T, b, n)
 end
 nlevelstate(b::NLevelBasis, n::Integer) = nlevelstate(ComplexF64, b, n)
+
+"""
+    paulix([T=ComplexF64,] b::NLevelBasis, pow=1)
+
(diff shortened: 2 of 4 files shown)


Krastanov:
Shouldn't these be methods of sigmax, sigmay, sigmaz instead? Should we implement them for FockBasis as well? If they are not methods of sigmax/y/z, should we implement them for SpinBasis(1//2)?

akirakyle:
There isn't a unique way to generalize the pauli operators as it depends on the properties one wants to preserve. [Wikipedia has a nice overview.](https://en.wikipedia.org/wiki/Generalizations_of_Pauli_matrices)

> Shouldn't these be methods of sigmax, sigmay, sigmaz instead?

Their implementation already generalizes them to arbitrary spin as the $J_x, J_y, J_z$ angular momentum operators which is one way to generalize the pauli operators that preserves them being hermitian. The clock and shift matrices generalize them being unitary.

> Should we implement them for FockBasis as well?

I don't think so, since the generalization of clock and shift to a harmonic oscillator gives the cannonical position and momentum operators. In fact, historically it went the other way: Weyl used these matrices to give finite dimensional systems the same commutator as that of position and momentum in an oscillator.

> If they are not methods of sigmax/y/z, should we implement them for SpinBasis(1//2)?

Similarly to above, I would worry a bit about confusion. In my mind, the whole difference between a spin system and an N level system is that the angular momentum operators are actual physical operations.

Krastanov:
Could you add commentary to this effect in the docstring? And a `see also [sigmax](@ref)...` and so on. So that the user can have an idea of why these names were chosen, etc. Including mentioning that these are also called "clock" and "shift" matrices, etc.

Krastanov (changes requested):
just adding a proper review to make my review queue a bit more organized

This is a great addition, thank you! To make it easier to navigate the difference between `sigma*` and `pauli*` could you add something based on your comments to the docstrings and add a `see also ` that crossreferences between them.

We should also double check we are not breaking the documentation build now that we are exporting or crossreferencing new things.

Krastanov:
Thanks for the additions! Whenever you feel the new docs are ready, please feel free to merge / bump version / release.

PR #201, opened 2025-05-29
Switch to TestItems

ErikQQY:
Part of https://github.com/qojulia/QuantumOptics.jl/issues/443


Krastanov (changes requested):
Thanks! I missed this while reviewing the related PR, my bad!

Same comments that I gave there apply here as well.

Krastanov:
One more thing here -- the JET tests have a bunch of conditionals in them that can be done more simply. Look into the QuantumClifford library for a reference on how to update the runtests.jl file.

ErikQQY:
> One more thing here -- the JET tests have a bunch of conditionals in them that can be done more simply.

I made the JET testing more formal by just using `test_package` to do static analysis, and fixed several bugs.

test/test_jet.jl
```diff
@@ -1,23 +1,6 @@
...
-                AnyFrameModule(Strided),
-                AnyFrameModule(StridedViews),
-                AnyFrameModule(Dates),
-                AnyFrameModule(SparseArrays))
-            )
-        @show rep
-        @test length(JET.get_reports(rep)) <= 24
-        @test_broken length(JET.get_reports(rep)) == 0
-    end
+    JET.test_package(QuantumOpticsBase, target_defined_modules = true)
```
Krastanov, line 5: great! Thanks!

test/runtests.jl
```diff
@@ -1,62 +1,5 @@
...
-if length(unavailable_tests) != 0
-    error("The following tests could not be found:\n", join(unavailable_tests, "\n"))
-end
-
-for name=names
-    if startswith(name, "test_") && endswith(name, ".jl")
-        include(name)
-    end
-end
+testfilter = ti->(contains(ti.filename, "test_") && endswith(ti.filename, ".jl") )
```
Krastanov, line 4: the filter is not necessary

src/manybody.jl
```diff
@@ -271,7 +271,7 @@ function manybodyoperator_2(mb::ManyBodyBasis, op::SparseOpType)
     Js = Int[]
     Vs = ComplexF64[]
     buffer = allocate_buffer(mb)
-    @inbounds for (row, column, value) in zip(findnz(op.data)...)
+    @inbounds for (row, column, value) in zip(findnz(sparse(op.data))...)
```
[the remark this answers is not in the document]
ErikQQY: I believe `op.data` should always be a sparse matrix, because the `data` in `op::SparseOpType` is always a sparse matrix, but the error case is that when the `data` in `op::SparseOpType` becomes an adjoint sparse matrix, for example:

```julia
julia> using SparesArrays, LinearAlgebra
julia> a = sprand(3, 3, 0.3)
3×3 SparseMatrixCSC{Float64, Int64} with 4 stored entries:
 0.493387   ⋅        0.925132
  ⋅        0.696328   ⋅ 
  ⋅         ⋅        0.0164289

julia> findnz(a)
([1, 2, 1, 3], [1, 2, 3, 3], [0.4933869378179172, 0.6963280350033298, 0.9251323625016207, 0.016428929253714863])

julia> findnz(a')
ERROR: MethodError: no method matching findnz(::LinearAlgebra.Adjoint{Float64, SparseMatrixCSC{Float64, Int64}})
The function `findnz` exists, but no method is defined for this combination of argument types.
```

I am a bit confused about why the previous tests didn't catch this, but when I was running tests locally, the error did happen.

PR #216, opened 2026-02-02
Makie bloch sphere extension 

Aniket-umass:
Some rendering examples:
Arbitrary:
bloch_arbitrary
Spindown:
bloch_spindown
Spinup:
bloch_spinup

Adds a Bloch sphere visualization for pure qubit states (Ket) via a Makie extension. The extension loads automatically when the user loads any Makie backend (CairoMakie or GLMakie) — no extra dependencies are added to the base package.

How it works
A pure qubit state |ψ⟩ = α|0⟩ + β|1⟩ is mapped to a point on the unit sphere via the Pauli expectation values: x = 2Re(ᾱβ), y = 2Im(ᾱβ), z = |α|² − |β|². The arrow points to the state on the sphere surface.
Customizable attributes

```
arrowcolor (default :red) — color of the state vector arrow
spherecolor (default :lightblue) — color of the sphere surface
spherealpha (default 0.15) — transparency of the sphere
showwireframe (default true) — show equator and meridian circles
showaxes (default true) — show dashed x/y/z axis lines
showlabels (default true) — show pole labels |0⟩ |1⟩ |+⟩ |-⟩ |+i⟩ |-i⟩
labelsize (default 18) — font size for pole labels
```

Files changed
ext/QuantumOpticsBaseMakieExt.jl — Makie recipe and blochsphere convenience constructor
src/visualization.jl — docstring and function stub for blochsphere
src/QuantumOpticsBase.jl — export and include for visualization.jl
Project.toml — Makie added as weak dependency
test/test_plotting.jl — tests using @testitem / TestItemRunner


ext/QuantumOpticsBaseMakieExt.jl
```diff
@@ -0,0 +1,60 @@
+module QuantumOpticsBaseMakieExt
+
+import QuantumOpticsBase: blochsphere, Ket
+using Makie
+using GeometryBasics
```
Krastanov: let's turn these into more specific ones. E.g. `using Makie: Makie, <only the list of things you actually need>` or using `Makie.whatever_you_need` in the code below. That way we can keep the namespaces in the codebase a bit cleaner.

ext/QuantumOpticsBaseMakieExt.jl
```diff
@@ -0,0 +1,60 @@
...
+# Configure axis visibility safely
+function configure_axis!(ax, showaxes::Bool)
+    if !showaxes
+        hidexdecorations!(ax)
+        hideydecorations!(ax)
+        hidezdecorations!(ax)
+    end
+end
+
+function blochsphere(state::Ket; arrowcolor=:red, spherealpha=0.35, showaxes=true)
```
Krastanov: This should probably be converted to a "recipe". Otherwise the majority of the capabilities of the plotting library will not be available (e.g. interactive fast updates to the plot and animations). See https://docs.makie.org/stable/explanations/recipes

Project.toml
```diff
@@ -8,15 +8,24 @@ FFTW = "7a1cc6ca-52ef-59f5-83cd-3a7055c09341"
...
+Makie = "ee78f7c6-11fb-53f2-987a-cfe4a2b5a57a"
 QuantumInterface = "5717a53b-5d69-4fa3-b976-0bf2f97ca1e5"
 Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
 RecursiveArrayTools = "731186ca-8d62-57ce-b412-fbd966d074cd"
 SparseArrays = "2f01184e-e22b-5df5-ae63-d93ebab69eaf"
 Strided = "5e0ebb24-38b0-5f93-81fe-25c709ecae67"
 UnsafeArrays = "c4a57d5a-5b31-53a6-b365-19f8c011fbd6"
 
+[weakdeps]
+CairoMakie = "13f3f980-e62b-5c42-98c6-ff1f3baf88f0"
```
Krastanov: Why is CairoMakie a weak dep?

Krastanov: you can just remove it, same for GLMakie

Krastanov (changes requested):
Thanks, this is a good start!

Check the comments I left in.

A few other things:
- this new functionality needs tests (check PBCCompiler.jl for an example of how to structure the tests)
- and it needs docs
- for ease of review, it would be helpful to post in the comments what a rendering looks like so we can discuss it before a full in-depth review.

Krastanov:
> `include("ext/QuantumOpticsBaseMakieExt.jl")`

You should not include extensions manually. They are supposed to be loaded when the library is imported together with the weakdeps for the extension. Check out the details in the Pkg.jl manual.

~/.julia/dev/QuantumOpticsBase/Project.toml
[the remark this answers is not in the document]
Krastanov: does not seem to be resolved

Project.toml
```diff
@@ -1,13 +1,14 @@
 name = "QuantumOpticsBase"
 uuid = "4f57444f-1401-5e15-980d-4471b28d5678"
 version = "0.5.9"
-
+
```
Krastanov: don't add "lint"

always check the file tab in the PR to look for unrelated changes that you have made by mistake

Project.toml
```diff
@@ -1,13 +1,14 @@
...
 version = "0.5.9"
-
+ 
 [deps]
 Adapt = "79e6a3ab-5dfb-504d-930d-738a2a938a0e"
 FFTW = "7a1cc6ca-52ef-59f5-83cd-3a7055c09341"
 FastExpm = "7868e603-8603-432e-a1a1-694bd70b01f2"
 FastGaussQuadrature = "442a2c76-b920-505d-bb47-c5924d526838"
 FillArrays = "1a297f60-69ca-5386-bcde-b61e274b549b"
+GeometryBasics = "5c1252a2-5f33-56bf-86c9-59e7332b4326"
```
Krastanov: you have added a new dependency -- this is supposed to be an extension

src/QuantumOpticsBase.jl
```diff
@@ -70,10 +70,18 @@ export Basis, GenericBasis, CompositeBasis, basis,
...
-        qfunc, wigner, coherentspinstate, qfuncsu2, wignersu2
+        qfunc, wigner, coherentspinstate, qfuncsu2, wignersu2, blochsphere
         #apply
                 apply!
 
+                ###########  Bloch Sphere API Stub  ###########
+
+function blochsphere(state)
+error("Blochsphere visualization requires a Makie extension.\n" *
+      "Load QuantumOpticsBase together with CairoMakie or GLMakie.")
```
Krastanov: instead of creating a method for a function, just declare a function without any methods

```
function blochsphere end
```

src/QuantumOpticsBase.jl
```diff
@@ -70,10 +70,18 @@ export Basis, GenericBasis, CompositeBasis, basis,
         PauliBasis, PauliTransferMatrix, DensePauliTransferMatrix,
                 ChiMatrix, DenseChiMatrix, avg_gate_fidelity,
         SumBasis, directsum, ⊕, LazyDirectSum, getblock, setblock!,
-        qfunc, wigner, coherentspinstate, qfuncsu2, wignersu2
+        qfunc, wigner, coherentspinstate, qfuncsu2, wignersu2, blochsphere
```
Krastanov: if you are exporting a function, it should be documented

also, notice that the exports are organized, so maybe add a comment header for this new category of functions, e.g. visualizations

src/QuantumOpticsBase.jl
```diff
@@ -70,10 +70,18 @@ export Basis, GenericBasis, CompositeBasis, basis,
...
         #apply
                 apply!
 
+                ###########  Bloch Sphere API Stub  ###########
```
Krastanov: do not make up new organization style comments

put this is a new file (notice that the exports above are organized by file) and include the file at the bottom

test/test_plotting.jl
```diff
@@ -0,0 +1,177 @@
...
+ 
+
+    @testset "blochsphere returns a Figure" begin
+        @test blochsphere(spinup(b)) isa Figure
+    end
+ 
+    @testset "blochsphereplot! returns a Plot" begin
+        fig = Figure()
+        ax  = Axis3(fig[1, 1])
+        @test blochsphereplot!(ax, spinup(b)) isa Makie.Plot
```
Krastanov: how is this passing if blochsphereplot is not exported?

test/test_plotting.jl
```diff
@@ -0,0 +1,177 @@
...
+        @test y ≈  0 atol=1e-10
+        @test z ≈  0 atol=1e-10
+    end
+ 
+    @testset "Bloch vector |-⟩ -x pole" begin
+        x, y, z = bloch_vector(normalize(spinup(b) - spindown(b)))
+        @test x ≈ -1 atol=1e-10
+        @test y ≈  0 atol=1e-10
+        @test z ≈  0 atol=1e-10
+    end
```
Krastanov: these tests seem to not test anything related to the library, rather some helper function defined in the test file itself

test/test_plotting.jl
```diff
@@ -0,0 +1,177 @@
...
+ 
+    @testset "Bloch vector |-i⟩ -y pole" begin
+        x, y, z = bloch_vector(normalize(spinup(b) - im*spindown(b)))
+        @test x ≈  0 atol=1e-10
+        @test y ≈ -1 atol=1e-10
+        @test z ≈  0 atol=1e-10
+    end
+
```
Krastanov: same with these -- what is the point of these tests

test/test_plotting.jl
```diff
@@ -0,0 +1,177 @@
...
+        end
+    end
+ 
+    @testset "Bloch vector has unit length for any pure state" begin
+        for (θ, φ) in [(0.0, 0.0), (π/4, π/3), (π/2, 0.0), (π/2, π), (π, 0.0)]
+            ψ = cos(θ/2)*spinup(b) + exp(im*φ)*sin(θ/2)*spindown(b)
+            x, y, z = bloch_vector(ψ)
+            @test x^2 + y^2 + z^2 ≈ 1 atol=1e-10
+        end
+    end
```
Krastanov: same issue -- what is the point of these tests

test/test_plotting.jl
```diff
@@ -0,0 +1,177 @@
...
+        @test isfile("test_bloch_plusi.png")
+        rm("test_bloch_plusi.png")
+    end
+ 
+    @testset "|-i⟩ renders without error" begin
+        fig = blochsphere(normalize(spinup(b) - im*spindown(b)))
+        save("test_bloch_minusi.png", fig)
+        @test isfile("test_bloch_minusi.png")
+        rm("test_bloch_minusi.png")
+    end
```
Krastanov: just pick one random arbitrary unnamed state and test for that -- these repeated tests do not really add much value to the test runner

ext/QuantumOpticsBaseMakieExt.jl
```diff
@@ -0,0 +1,157 @@
...
+        ygridvisible       = false,
+        zgridvisible       = false,
+        xypanelvisible     = false,
+        xzpanelvisible     = false,
+        yzpanelvisible     = false,
+    )
+    lim = Float32(get(kwargs, :lim, 1.6))
+    Makie.limits!(ax, -lim, lim, -lim, lim, -lim, lim)
+    blochsphereplot!(ax, state; kwargs...)
+    return fig
```
Krastanov: return also the axis and the plot object

ext/QuantumOpticsBaseMakieExt.jl
```diff
@@ -0,0 +1,157 @@
...
+        yspinesvisible     = false,
+        zspinesvisible     = false,
+        xgridvisible       = false,
+        ygridvisible       = false,
+        zgridvisible       = false,
+        xypanelvisible     = false,
+        xzpanelvisible     = false,
+        yzpanelvisible     = false,
+    )
+    lim = Float32(get(kwargs, :lim, 1.6))
```
Krastanov: this seems like a non-standard name -- pick a name that matches the Makie API

ext/QuantumOpticsBaseMakieExt.jl
```diff
@@ -0,0 +1,157 @@
...
+        shaftradius   = 0.018,
+        tipradius     = 0.050,
+        tiplength     = 0.10,
+        lim           = 1.6,
+    )
+end
+ 
+function Makie.convert_arguments(::Type{<:Makie.Plot{blochsphereplot}}, state::Ket)
+    return (state,)
+end
```
Krastanov: Why is this necessary?

ext/QuantumOpticsBaseMakieExt.jl
```diff
@@ -0,0 +1,157 @@
...
+        spherealpha   = 0.15,        
+        showwireframe = true,        
+        showaxes      = true,        
+        showlabels    = true,        
+        labelsize     = 18,
+        title         = "Bloch Sphere",
+        shaftradius   = 0.018,
+        tipradius     = 0.050,
+        tiplength     = 0.10,
+        lim           = 1.6,
```
Krastanov: lim does not seem like a standard name, verify it is correct

a bunch of these do not seem to be used (e.g. title) -- just remove them

Krastanov (changes requested):
looks like it is on a good track

could you address comments, resolve the merge conflict, verify that tests actually pass in CI, and post a few example plots in the PR description so that we know how it is being rendered

ext/QuantumOpticsBaseMakieExt.jl
```diff
@@ -0,0 +1,149 @@
...
+        )
+            text!(p, pos; text = lbl, fontsize = ls, align = align)
+        end
+    end
+
+    return p
+end
+
+
+function QuantumOpticsBase.blochsphere(state::Ket; limits=1.6, kwargs...)
```
Krastanov: The API here is a bit unidiomatic.

Let's focus first on having the default Makie recipe functions be in the QuantumOpticsBase namespace. See https://docs.makie.org/stable/explanations/recipes#Makie-Package-Extension

Then, after we have the default functions, we can provide an extra API that is not idiomatic to Makie that provides not just a "plot" but also "axis" manipulation.

Krastanov: see https://github.com/QuantumSavory/QuantumSavory.jl/blob/master/src/plots.jl, in particular `registernetplot_axis`

Krastanov: rename to blochsphereplot_axis as in: https://github.com/QuantumSavory/QuantumSavory.jl/blob/766f69c7f5e595328d46883fb3bd3f945e373b0d/ext/QuantumSavoryMakie/QuantumSavoryMakie.jl#L340-L348

.gitignore
```diff
@@ -1,3 +1,4 @@
 /docs/build/
 Manifest.toml
 .vscode
+~/.julia/
```
Krastanov: a `tilda` inside of a gitignore does not really make sense to me -- if you need it, you probably should reorganize how you have set up your work environment

test/test_plotting.jl
Krastanov, line 1: make sure that plotting of observables and updating these observables also works

Krastanov (changes requested):

Krastanov:
Use error hints so that the user is told that they need to import a makie backend. See here https://github.com/QuantumSavory/QuantumSavory.jl/blob/766f69c7f5e595328d46883fb3bd3f945e373b0d/src/QuantumSavory.jl#L103

Krastanov:
> Use error hints so that the user is told that they need to import a makie backend. See here https://github.com/QuantumSavory/QuantumSavory.jl/blob/766f69c7f5e595328d46883fb3bd3f945e373b0d/src/QuantumSavory.jl#L103

Do not forget to do this, probably in a future PR

PR #224, opened 2026-04-23
bump RecursiveArrayTools and cleanup for CI

Krastanov:


Krastanov:
the downgrade failure is a Reexport issue that will need to be fixed in the test dependency QuantumOptics.jl

PR #227, opened 2026-06-03
Add Wigner function computation and Makie visualization extension

Aniket-umass:
Plot examples:
**Coherent state:**  
wigner_coherent

**Fock state:**
wigner_fock3

Adds a Makie recipe for visualizing the Wigner quasi-probability distribution of quantum states as a 2D heatmap, following the same extension pattern as the existing Bloch sphere visualization.
Changes

**ext/QuantumOpticsBaseMakieExt.jl:** Adds WignerPlot recipe producing a heatmap with a diverging colormap (:RdBu) pinned to zero, so positive regions (blue) and negative regions (red) are visually distinct. Supports reactive 
Observable state updates.

**src/visualization.jl:** Adds function stubs for wignerplot, wignerplot!, and wignerplot_axis.

**src/QuantumOpticsBase.jl:** Exports the three new visualization functions.

**test/test_wigner_plotting.jl:** Tests for return types, coherent/Fock state rendering, custom attributes (xrange, prange, npoints, colormap), Observable reactivity, and error handling for unsupported basis types.

**test/Project.toml:** Adds Makie as a test dependency (needed for using Makie: Observable in both Bloch sphere and Wigner tests).


Krastanov:
FYI, currently this PR seems to be deleting what was added in the previous PR. Something was probably messed up when resolving the merge conflict

Aniket-umass:
> FYI, currently this PR seems to be deleting what was added in the previous PR. Something was probably messed up when resolving the merge conflict

Alright, I will look into it

Aniket-umass:
All Wigner and Bloch sphere plotting tests pass (9/9 each). The remaining CI failures (sciml_broadcast_interfaces, aqua, apply, Documentation, Downgrade) are pre-existing on the master branch.

test/test_wigner_plotting.jl
```diff
@@ -0,0 +1,69 @@
...
+        @test isfile("test_wigner_observable.png")
+        rm("test_wigner_observable.png")
+    end
+
+    # ── Error handling ────────────────────────────────────────────────────────
+    @testset "Wrong basis type throws error" begin
+        # wigner is only defined for FockBasis states
+        b_spin = SpinBasis(1//2)
+        ψ_spin = spinup(b_spin)
+        @test_throws Exception wignerplot_axis(ψ_spin)
```
Krastanov: this test checks for ANY exception, so it is not actually validating that you are getting the exception you are expecting. `@test_throws` supports using a substring of the error message as a quick way to provide more specificity, use that instead.

Krastanov (changes requested):
Looks good, up to minor test clarification. Still outstanding need to make error hints that point out the need for Makie backends so that the method errors are more legible.

Krastanov (changes requested):
the way the error messages are done is a step backwards -- use the approach we had discussed with error hints

src/QuantumOpticsBase.jl
```diff
@@ -107,4 +108,15 @@ include("printing.jl")
 include("apply.jl")
 include("visualization.jl")
 
+function __init__()
```
Krastanov, line 114: Use WeakDepHelpers.jl, do not reimplement it.

Krastanov (commented):
this looks good, on merging of 238 there will probably be merge conflicts, otherwise same comment as https://github.com/qojulia/QuantumOpticsBase.jl/pull/238#pullrequestreview-4655101003

looks good but before we merge:

-    we need the release of weakdephelpers for 1.10
-    we need to see passing tests (all of them, up to some unrelated issues with sciml right now)

Krastanov:
Thanks, @Aniket-umass , this was merged after some minor rework in #268

PR #238, opened 2026-06-23
Add Fock distribution bar chart visualization

Aniket-umass:
## Fock distribution bar chart visualization

Adds a `fockdistributionplot` recipe to the Makie extension, visualizing the
Fock-state (number-state) occupation probabilities P(n) of a quantum state as a
bar chart. This mirrors QuTiP's `plot_fock_distribution` and complements the
existing Bloch sphere (spin) and Wigner (phase-space) plotters, extending
visualization coverage to the Fock basis.

### What it does
- For a `Ket`, plots `P(n) = |⟨n|ψ⟩|²`.
- For a density operator, plots the diagonal `P(n) = ⟨n|ρ|n⟩`.
- Default styling matches QuTiP: translucent green bars, y-axis fixed to [0, 1]
  (toggle with `unit_y_range=false`), x = "Fock number", y = "Occupation probability".

### API
- `fockdistributionplot(state; kwargs...)` / `fockdistributionplot!(ax, state; kwargs...)`
- `fockdistributionplot_axis([ax,] state; kwargs...) -> (Figure, Axis, Plot)`
- Attributes: `color`, `alpha`, `width`; axis kwarg: `unit_y_range`.

When no Makie backend is loaded, calling these gives a normal `MethodError`
with a hint to load one (via `Base.Experimental.register_error_hint`).

### Tests
`test/test_fock_plotting.jl` (tag `:plotting`) covers return types, rendering of
coherent / number / thermal states, custom attributes, the `unit_y_range`
toggle, and `Observable` reactivity.

### Notes
- No `Project.toml` changes — reuses the Makie weakdep/extension already in place.

Example:-
Screenshot 2026-07-06 at 10 18 44 PM

PR #241, opened 2026-08-08
Add wave function visualization

Aniket-umass:
Wave function visualization

Adds a wavefunctionplot recipe to the Makie extension, drawing the wave function of a particle state against position or momentum. This complements the existing Bloch sphere (spin), Wigner (phase space) and Fock distribution (number basis) plotters, covering the remaining basis in which states are routinely inspected.

What it does
For a Ket in a PositionBasis, plots ψ(x) against x.
For a Ket in a MomentumBasis, plots ψ(p) against p, with the axis label following the basis.
part selects what is drawn: :abs2 (default, probability density), :abs, :real, :imag.

Amplitudes are divided by √Δx before plotting. gaussianstate and friends store ⟨xᵢ|ψ⟩ = √Δx·ψ(xᵢ) so that the discretised Ket is normalised; without dividing it out the plotted height would depend on the grid resolution rather than on the physical state.

Only pure states are accepted — a density operator has no wave function, so it raises rather than silently plotting something else.

API
wavefunctionplot(state; kwargs...) / wavefunctionplot!(ax, state; kwargs...)
wavefunctionplot_axis([ax,] state; kwargs...) -> (Figure, Axis, Plot)
Attribute: part.

Parts are composable rather than bundled, so components can be overlaid on one axis without the recipe hardcoding any styling:

julia
fig, ax, _ = wavefunctionplot_axis(ψ; part=:real)
wavefunctionplot!(ax, ψ; part=:imag)

Known limitation: overlaid parts currently draw in the same colour, because the recipe does not yet declare cycle = [:color]. Happy to add that here or as a follow-up — I did not want to guess at the Makie 0.24 attribute-inheritance API without checking it properly.

When no Makie backend is loaded, calling these gives a MethodError with a hint to load one, via the same WeakDepHelpers mechanism used in #227 and #238.

Tests

test/test_wavefunction_plotting.jl (tag :plotting) covers return types, rendering in both position and momentum bases, every part, overlaying parts on one axis, Observable reactivity, and the three error paths (wrong basis, mixed state, unknown part).

Note on CI

Two failures on arrival, neither from this code:

Julia 1.10 — this PR uses WeakDepHelpers, so it hits the same unsatisfiable resolve as #227 and #238 until WeakDepHelpers 0.1.2 is registered (QuantumSavory/WeakDepHelpers.jl#13). The registry currently holds only 0.1.0 and 0.1.1, both pinned julia = "1.12.0 - 1".
Julia 1.12 — test_sciml_broadcast_interfaces.jl errors with MethodError: no method matching iterate(::Ket{...}) raised inside OrdinaryDiffEqCore's _ode_initdt_iip. This is repo-wide and unrelated.

test_wavefunction_plotting.jl passes 14/14 locally on Julia 1.12.6, the same version CI runs, alongside test_plotting.jl 9/9.

wavefunction_example


Krastanov:
This looks pretty nice on first quick review. Something that confuses me quite a bit though is, why is the line yellow? That is not the default in Makie and I do not see where it is overwriten here

Krastanov:
What does the following mean, could you elaborate (on all clauses):

> Amplitudes are divided by √Δx before plotting. gaussianstate and friends store ⟨xᵢ|ψ⟩ = √Δx·ψ(xᵢ) so that the discretised Ket is normalised; without dividing it out the plotted height would depend on the grid resolution rather than on the physical state.

Krastanov:
The plot shows two lines but the y axis has just the label "Re(psi)". What are the two lines, how are they to be distinguished, what is the meaning of the y axis label?

Krastanov:
Thanks, @Aniket-umass , this was merged after some minor rework in #268

PR #248, opened 2026-08-16
Add cold-start latency benchmark

Krastanov-agent:
## Summary

- add a cold-start latency workflow based on QuantumSavory.jl's precompilation benchmark
- compare the exact pull-request base and head with Julia 1.12.6, two cache builds, and five samples per build
- measure package-cache build size/time, import time, first and warm Fock/composite calls, compilation, and total latency
- add deterministic, self-checking scenarios and upload the raw measurements with a descriptive job summary
- add the `PrecompileTools` dependency and an empty precompilation scaffold so later workload PRs use identical dependency metadata

The design follows [QuantumSavory.jl#529](https://github.com/QuantumSavory/QuantumSavory.jl/pull/529) as closely as possible. The harness uses a shared resolved consumer manifest, fresh overlay depots, fixed single-thread execution, counterbalanced build order, and offline measurement after setup.

The one-time comparison against current `master` has an exact exception for the two new `PrecompileTools` metadata entries. Such a bootstrap run is marked non-reportable and resolves from the verified dependency-superset candidate. Subsequent precompilation PRs have identical package metadata and do not use this exception.

This is the first of three stacked changes. Follow-up PRs add representative example workloads, then residual directives selected from `--trace-compile-timing` observations strictly longer than 50 ms.

## Validation

- `Pkg.test()` on Julia 1.12.6: 2,130 passed, 2 expected broken
- all ten benchmark scenarios in fresh Julia 1.12.6 processes
- representative scenarios on Julia 1.10.11
- `QOB_PRECOMPILE_TRACE=compile` and `dispatch` diagnostics on Julia 1.12.6
- current-HEAD end-to-end harness smoke test with one build and one sample
- reverse bootstrap metadata comparison rejected as expected
- shell, Julia syntax, YAML, and whitespace checks


Krastanov-agent:
CI note (2026-08-16): the existing latest-Julia and downgrade jobs now resolve Julia 1.12.7 and fail the six exact floating-point comparisons at `test/test_superoperators.jl:174-179`. I reproduced the same six failures on untouched upstream commit `18dc5ef` and on this branch, with identical maximum element differences from `5.55e-17` to `1.78e-15`. On untouched upstream with Julia 1.12.6, all six comparisons are exactly equal. The new cold-start latency job passes. I have not mixed an unrelated Julia 1.12.7 test-tolerance change into this PR.

